Issue #4, opened 2017-05-16
Setup basic extension


jonathanKingston:
- Steal existing icon
- Make manifest file
- Make a browser action to create a tab

princiya:
This is what I created as part of my application process to Outreachy: https://github.com/princiya/log-network-requests
A simple web extension

biancadanforth:
Thanks Princiya! Yes I made [one](https://github.com/biancadanforth/network-requests-webext) too, and I have [Jonathan](https://github.com/jonathanKingston/extension-debugging)'s as a reference as well. No shortage of resources on this front, heh.

PR #8, opened 2017-05-16
Set up basic web extension

biancadanforth:
In response to [Issue #4](https://github.com/pauljt/lightbeam-we/issues/4) 
Note: The `filter()` method is used with `tabs.query()` instead of a match pattern or regular expression to check if Lightbeam is already open. This is because Mozilla match patterns do not support any URLs with schemes other than: "http", "https", "file", "ftp", "app". This webext has a scheme of "moz-extension". It also doesn't support regular expressions in place of match patterns.


background.js
```diff
@@ -0,0 +1,34 @@
+'use strict';
+
+// browser for Firefox, chrome for Chrome
+if (!browser) {
+  const browser = chrome;
```
jonathanKingston: Drop this, we will add a shim later to fix this.

background.js
```diff
@@ -0,0 +1,34 @@
+'use strict';
```
jonathanKingston: Yay!

manifest.json
```diff
@@ -0,0 +1,29 @@
...
+  "homepage_url": "https://github.com/pauljt/lightbeam-we",
+
+  "icons": {
+    "48": "images/lightbeam-48.png",
+    "96": "images/lightbeam-96.png"
+  },
+
+  "permissions": [
+  	"webRequest",
+  	"<all_urls>",
```
jonathanKingston: Nit: indenting

index.html
```diff
@@ -0,0 +1,17 @@
...
+    <title>Lightbeam</title>
+    <meta name="description" content="Lightbeam is a Firefox add-on that uses interactive visualizations to show you the relationships between third parties and the sites you visit.">
+    <meta name="viewport" content="width=device-width, initial-scale=1">
+    <!-- <link rel="stylesheet" href="style.css"> -->
+    <link rel="shortcut icon" href="images/lightbeam-48.png" type="image/x-icon">
+    <link rel="icon" href="images/lightbeam-48.png" type="image/x-icon">
+  </head>
+  <body>
+  <h1>Lightbeam</h1>
+  <!-- <script src="script.js"></script> -->
```
jonathanKingston: As we will likely need this it is fine, code comments are pretty bad. (don't follow my example when I do it :D)

background.js
```diff
@@ -0,0 +1,34 @@
...
+	}
+
+	// Only open a new Lightbeam instance if one isn't already open.
+	const lightbeamTab = await isOpen();
+	if (!lightbeamTab) {
+		browser.tabs.create(
+			{url: 'index.html'},
+			(tab) => {
+				const lightbeamTab = tab.id;
+				console.log(lightbeamTab);
```
jonathanKingston: nit: remove debugging

background.js
```diff
@@ -0,0 +1,34 @@
...
+			}
+		);	
+	}
```
jonathanKingston: File a follow up bug for focusing existing tabs

background.js
```diff
@@ -0,0 +1,34 @@
...
+}
```
jonathanKingston: Also we will need to decide on a code style and agree on it. As this is yours and @princiya's project we should file a bug and get an ESLint config going. Personally my style is different but that is fine :)

manifest.json
jonathanKingston: V doesn't match ^

PR #16, opened 2017-05-23
Simplified data capture to storage API. Fixes #7.

### LINKED ISSUE #7: Initial capture of data
In #1 we set out to decide what is needed to capture.

In the MVP we should aim to have a class/object that can be setup to manage the communication with the browser then use the storage class.

biancadanforth:
Makes data capture agnostic of storage object/method.

### DIFF
--- a/background.js
+++ b/background.js
@@ -3,8 +3,9 @@
 // When the user clicks browserAction icon in toolbar, execute runLightbeam function
 browser.browserAction.onClicked.addListener(runLightbeam);
 
-async function runLightbeam() {
+capture.init();
 
+async function runLightbeam() {
   // Checks to see if Lightbeam is already open. Returns true if it is, false if not.
   async function isOpen() {
     const tabs = await browser.tabs.query({});
--- a/capture.js
+++ b/capture.js
@@ -1,0 +1,46 @@
+const capture = {
+  init() {
+    this.addListeners();
+  },
+
+  addListeners() {
+    // listen for each HTTP response
+    browser.webRequest.onResponseStarted.addListener(this.sendThirdParty, {urls: ['<all_urls>']}
+    );
+    // listen for tab updates
+    browser.tabs.onUpdated.addListener(this.sendFirstParty);
+  },
+
+  // capture third party requests
+  async sendThirdParty(response) {
+    const tab = await browser.tabs.get(response.tabId);
+    let tabUrl = new URL(tab.url);
+    let targetUrl = new URL(response.url);
+    let originUrl = new URL(response.originUrl);
+
+    if (targetUrl !== tabUrl) {
+      const thirdPartyData = {
+        document: tabUrl.hostname,
+        target: targetUrl.hostname,
+        origin: originUrl.hostname,
+        requestTime: response.timeStamp
+      };
+      console.log('storage.thirdPartyRequest:', tabUrl, thirdPartyData);
+    }
+  },
+
+  // capture first party requests
+  sendFirstParty(tabId, changeInfo, tab) {
+    const tabUrl = new URL(tab.url);
+    // ignore about:* pages and non-visible tabs
+    if (tab.status === 'complete'
+      && tabUrl.protocol !== 'about:'
+      && tabId !== browser.tabs.TAB_ID_NONE)
+       {
+      const firstPartyData = {
+        faviconUrl: tab.favIconUrl,
+      };
+      console.log('storage.firstPartyRequest:', tabUrl.hostname, firstPartyData);
+    }
+  }
+};
--- a/manifest.json
+++ b/manifest.json
@@ -23,6 +23,9 @@
   },
 
   "background": {
-    "scripts": ["background.js"]
+    "scripts": [
+      "capture.js",
+      "background.js"
+    ]
   }
 }


captureData.js
```diff
@@ -0,0 +1,47 @@
...
+        target: targetUrl,
+        origin: originUrl,
+        requestTime: response.timeStamp
+      };
+      console.log('storage.thirdPartyRequest:', tabUrl, thirdPartyData);
+    }
+  }
+
+  // Any subdomains can cause redundancy in storage. Returns the URL with no subdomains.
+  function removeSubdomain(url) {
```
jonathanKingston: Lightbeam currently shows hostnames?

captureData.js
```diff
@@ -0,0 +1,47 @@
+function captureData() {
```
jonathanKingston: We likely want this as an object too, shall I approve like this and we can clean up?

```
const captureData {
  init() {
    ...
  },
  otherMethod() {
  }
}
```

captureData.js
princiya: it will be good to move such functions to utils.js file?

captureData.js
```diff
@@ -0,0 +1,47 @@
...
+    let targetUrl = new URL(response.url);
+    targetUrl = removeSubdomain(targetUrl.hostname);
+    let originUrl = new URL(response.originUrl);
+    originUrl = removeSubdomain(originUrl.hostname);
+
+    // capture first party requests
+    const firstPartyData = {
+      faviconUrl: tab.favIconUrl,
+      // access determined by storage
+      firstAccess: 0,
```
jonathanKingston: We can delete these keys

captureData.js
jonathanKingston: This also won't account for Left to Right domains... it's a nasty problem we likely won't want to solve.

jonathanKingston: This also won't work for .co.uk will it?

jonathanKingston:
I think this is great, remove the firstAccess keys and the clean method as I don't think we need this.

To my knowledge Lightbeam doesn't group by domain name nor is there a web API to sanely get that.

captureData.js
```diff
@@ -0,0 +1,47 @@
+function captureData() {
+
+  browser.webRequest.onResponseStarted.addListener(collectRequests,
+        {urls: ['<all_urls>']}
```
jonathanKingston: Inconsistent indent.

captureData.js
biancadanforth: As we discussed in today's meeting, I will just remove this function; which doesn't really work the way I intend it to anyway.

background.js
```diff
@@ -4,7 +4,7 @@
 browser.browserAction.onClicked.addListener(runLightbeam);
 
 async function runLightbeam() {
-
+  captureData();
```
groovecoder: This prevents the add-on from capturing any data unless the browserAction icon is clicked. Is that what we want?

captureData.js
biancadanforth: I will restructure!

background.js
groovecoder: Because this is in the `runLightbeam` function, which is only called by `browserAction.onClicked`, no data is captured unless/until the user clicks the `browserAction` icon. Is that what we want?

biancadanforth: What a polite way of giving feedback! I have considered it and moved it outside of `runLightbeam`.

capture.js
```diff
@@ -0,0 +1,53 @@
...
+        document: tabUrl,
+        target: targetUrl,
+        origin: originUrl,
+        requestTime: response.timeStamp
+      };
+      console.log('storage.thirdPartyRequest:', tabUrl, thirdPartyData);
+    }
+  },
+
+  checkNewSite(tabId, changeInfo, tab) {
```
jonathanKingston: File a follow up to prevent non visible frames from showing. tabId < 0 to my memory.

capture.js
```diff
@@ -0,0 +1,53 @@
...
+    if (changeInfo.url) {
+      this.nextSite = changeInfo.url;
+    }
+    //favicon not available until tab status is complete; opening a new tab (about:blank) does not trigger a call to storage API
+    if ((this.nextSite !== this.prevSite) && (changeInfo.status === 'complete') && (this.nextSite !== null) && (this.nextSite !== 'about:blank')) {
```
jonathanKingston: File a followup to ignore all `about:` urls.

jonathanKingston: Code style nit, line length is likely wider than 80 chars these days but I would break by statement here:
```
 if (this.nextSite !== this.prevSite
     && changeInfo.status === 'complete'
     && this.nextSite !== null
     && this.nextSite !== 'about:blank') {
```

jonathanKingston: I'm not really sure on the rationale to prevent multiple calls of a complete page load of the same first party. Sure this will over call into the storage, however it is also storing state internally with the added side effect that the first call a tab makes looks like it would be ignored?

Could you explain why `prevSite` and `nextSite` exist?

jonathanKingston: I'm not 100% certain that this API will provide a url always as it seems to be about state tracking of tab updates. (The MDN docs aren't actually clear or leave it to ambiguity which suggest anything optional could be so).

Because all the [changeInfo](https://developer.mozilla.org/en-US/Add-ons/WebExtensions/API/tabs/onUpdated#changeInfo) is optional I would only be checking for changes when completed:

```js
async collectRequests(tabId, updated) {
  const tab = await = browser.tabs.get(tabId);
  const tabUrl = new URL(tab.url);
  if (updated.status === "completed") {
    console.log('storage.firstPartyUpdate:', tabUrl.hostname, tab);
  }
})
```

My understanding is the API will only provide urls when they have changed which means the storage should take care of piecemeal building up what a tab represents.

capture.js
jonathanKingston: `browser.tabs.TAB_ID_NONE` is actually the correct way to do this.

capture.js
biancadanforth: The first call a tab makes isn't ignored. I have tested it. This is thanks to `this.nextSite` and `this.prevSite`. I initialize `this.prevSite` to `null` and when the tab updates and status is complete, that differs from the tab's new URL and it calls the storage API.
The `this.nextSite` and `this.prevSite` are the way I know whether the request represents a new first party request (so I don't make a bunch of calls to storage needlessly). This won't prevent complete redundancy, but it helps a lot. It sounds like you prefer storage to handle that. Why is that better? I'm perfectly okay changing it, just want to make sure I am following the reasoning.

biancadanforth: To be clear here: you are suggesting I remove the `checkNewSite` function, and instead only listen for the `tab.status === 'complete'` inside of the `collectRequests` method?

capture.js
biancadanforth: Per Google's own [documentation ](https://chromium.googlesource.com/chromium/src/+/master/chrome/common/extensions/api/tabs.json), line 137, `browser.tabs.TAB_ID_NONE` is always -1. How does this help me to distinguish non-visible frames?

biancadanforth: Also, looking at my console and clicking around in `about:debugging`, neither it nor about:blank have made any HTTP requests, do some `about:*` pages make requests? Seems like if I don't make calls to the storage API except in cases of HTTP requests/responses, maybe I don't need to filter for these non-visible i.e. about:* tabs?
If I did, what about something like this:
```javascript
if (tab.status === 'complete' && !(re.test(tabUrl))) {
  const firstPartyData = {
    faviconUrl: tab.favIconUrl,
  };
  console.log('storage.firstPartyRequest:', tabUrl.hostname, firstPartyData);
}
```

capture.js
```diff
@@ -0,0 +1,47 @@
...
+    originUrl = originUrl.hostname;
+
+    //capture first party requests
+    const re = /^about:/;
+    // ignore non-visible tabs; i.e. about:* pages
+    if (tab.status === 'complete' && !(re.test(tabUrl))) {
+      const firstPartyData = {
+        faviconUrl: tab.favIconUrl,
+      };
+      console.log('storage.firstPartyRequest:', tabUrl, firstPartyData);
```
jonathanKingston: This will fire for any third party request right?

capture.js
```diff
@@ -0,0 +1,47 @@
...
+  async collectRequests(response) {
+    const tab = await browser.tabs.get(response.tabId);
+    let tabUrl = new URL(tab.url);
+    tabUrl = tabUrl.hostname;
+    let targetUrl = new URL(response.url);
+    targetUrl = targetUrl.hostname;
+    let originUrl = new URL(response.originUrl);
+    originUrl = originUrl.hostname;
+
+    //capture first party requests
```
jonathanKingston: This comment isn't relevant to the line below, was an if statement removed?

jonathanKingston: This likely needs to be checking for `response.type === "main_frame"` and third parties being the else clause

jonathanKingston: nit: space before `//` there is an ESLint rule for this to enforce consistency (when I say nit it is to be taken as, hey I'm pedantic and I'm sorry about that)

capture.js
```diff
@@ -0,0 +1,47 @@
...
+    // ignore non-visible tabs; i.e. about:* pages
+    if (tab.status === 'complete' && !(re.test(tabUrl))) {
+      const firstPartyData = {
+        faviconUrl: tab.favIconUrl,
+      };
+      console.log('storage.firstPartyRequest:', tabUrl, firstPartyData);
+    }
+
+    // capture third party requests
+    if (targetUrl !== originUrl) {
```
jonathanKingston: I need to check if this is true... I haven't tested this yet. originUrl is where it loads from rather than the top level frame URL from my testing. This would treat the first party being loaded by a third party as another third party I suspect.
So:

- google.com
  - other.com/tracking.js
    - google.com/callback.js << considered a third party

capture.js
```diff
@@ -0,0 +1,47 @@
...
+    const tab = await browser.tabs.get(response.tabId);
+    let tabUrl = new URL(tab.url);
+    tabUrl = tabUrl.hostname;
+    let targetUrl = new URL(response.url);
+    targetUrl = targetUrl.hostname;
+    let originUrl = new URL(response.originUrl);
+    originUrl = originUrl.hostname;
+
+    //capture first party requests
+    const re = /^about:/;
```
jonathanKingston: `URL(urlString).scheme === "about"` is likely cleaner however I do similar, just wanted to highlight :)

biancadanforth: @jonathanKingston `scheme` isn't a valid key on a URL object. Looks like the key is called `protocol`. Also, oddly, the scheme in this case would be `about:` (with colon)!

capture.js
biancadanforth: Yes I will update it to fire onUpdate

capture.js
biancadanforth: You're right. I have seen `originUrls` that don't match `tabUrls`. I will change this to `tabUrl`.

Issue #15, opened 2017-05-22
Structure of store object


princiya:
Based on our slack discussions, this is the proposed structure for the store object:
```
websites: {
    'website.com': {
      faviconUrl: '/link/to/url',
      firstAccess: date,
      lastAccess: date,
      thirdPartyRequests: {
        'a.com': {
          requestTime: date,
          cookies: []
        },
        'b.com': {}
        'c.com': {}
      }
    }
 }
```

biancadanforth:
Thanks! @princiya, @jkt I believe we also discussed not capturing cookies as part of the MVP? Looking at slack history.

biancadanforth:
I have removed cookies, replacing it with a new key `additionalRequests`. This key indicates any requests a third party makes to additional third parties. For example, when visiting youtube.com, doubleclick.net makes an additional request to 2mdn.net.
```javascript
websites: {
    'website.com': {
      faviconUrl: '/link/to/url',
      firstAccess: date,
      lastAccess: date,
      thirdPartyRequests: {
        'a.com': {
          requestTime: date,
          additionalRequests: []
        },
        'b.com': {}
        'c.com': {}
      }
    }
 }
```

biancadanforth:
I have removed `cookies` and added a `loadedBy` key to show additional requests made by third parties as the originUrl.

```javascript
websites: {
  'website.com': {
    faviconUrl: '/link/to/url',
    firstAccess: date,
    lastAccess: date,
    thirdPartyRequests: {
      'doubleclick.net': {
        requestTime: date,
      },
      '2mdn.net': {
        requestTime: date,
        loadedBy: 'doubleclick.net'
      },
      'b.com': {},
      'c.com': {}
    }
  }
 }
```

jonathanKingston:
It might not have been clear on this.

In https://github.com/electrolyfish/lightbeam-we/issues/7#issuecomment-303249670 I mentioned that the capture should be isolated from the storage code.

I'm also not exactly worried about how we store these right now. We can probably aggregate data sent to the storage object under "hostname" for both for first party and third party requests. This aggregating of the data could be done inside the storage code or the collation of the data code (for now it doesn't matter if storage receives hostnames or url).

First access and last access, should be implemented in the storage code based upon an access time passed to the request. The collation code shouldn't keep state of existing requests really.

[...]

Issue #6, opened 2017-05-16
Initial storage class


jonathanKingston:
Create a class with CRUD functionality (Create Read Update Delete) for pushing something into a storage database.

This will allow us to change the storage later whilst keeping the same wrapping class.

```
class {
  create() {
   // Some stuff
  }
  ...
}
```

jonathanKingston:
As mentioned on slack, it could be implemented as a class or instance object.
There would be a need for a singleton possibly if a class is used.

Feel free to use either or ask as and when you have issues :)

PR #17, opened 2017-05-23
Store object: get method

princiya:
Related issue: #6 
Structure of the store object: #15 

The `get()` method for the store object.


index.html
```diff
@@ -1,15 +1,15 @@
...
-    <meta name="description" content="Lightbeam is a Firefox add-on that uses interactive visualizations to show you the relationships between third parties and the sites you visit.">
-    <meta name="viewport" content="width=device-width, initial-scale=1">
-    <link rel="shortcut icon" href="images/lightbeam-48.png" type="image/x-icon">
-    <link rel="icon" href="images/lightbeam-48.png" type="image/x-icon">
-  </head>
-  <body>
-  <h1>Lightbeam</h1>
-  </body>
-</html>
+<!doctype html>
```
princiya: oops, I didn't change this file!

js/store.js
```diff
@@ -0,0 +1,77 @@
+const store = {
+  init() {
+    this.addListeners();
+  },
+
+  getAll() {
+    return this.get();
+  },
+
+  getFirstParties() {
```
princiya: do we really need this function?

js/store.js
```diff
@@ -0,0 +1,77 @@
...
+     * @return websites object
+     */
+    const websites = await browser.storage.local.get(params);
+    return websites || null;
+  },
+
+  async set(websites) {
+    /**
+      *@todo rewrite this method so that websites object is updated with additional hostnames
+    */
```
princiya: we need a local copy of websites, make sure new hostnames are added to it and then pass the whole websites object to the set method.

js/store.js
jonathanKingston: Yeah to build the graph I think we do... we can add later though :)

jonathanKingston: Maybe getAll is fine too

js/store.js
```diff
@@ -0,0 +1,77 @@
...
+
+  setFirstParty(param) {
+    /**
+     * @param object with hostname as key
+     * @todo validate the param
+     */
+    return this.set(param);
+  },
+
+  async get(params) {
```
jonathanKingston: This code probably belongs better in `getAll()` we can add this later if we need it.

index.html
jonathanKingston: This might be line endings perhaps? You can reset a file to a branch in future if this happens again. Lets ignore for now though.

js/store.js
```diff
@@ -0,0 +1,77 @@
...
+  },
+
+  async getThirdParties(hostname) {
+    if (!hostname) {
+      return null;
+    }
+
+    const storage = await this.get();
+
+    return storage.websites[`${hostname}`] ?
```
jonathanKingston: Use `storage.websites[hostname]` instead.
Multiline ternary statements like this are actually hard to read, I would personally go for something like:
```
const hostNameData = storage.websites[hostname];
if ("thirdPartyRequests" in hostNameData) {
  return hostNameData.thirdPartyRequests;
}
return null;
```

js/store.js
```diff
@@ -0,0 +1,77 @@
...
+      return null;
+    }
+
+    const storage = await this.get();
+    return storage.websites[`${hostname}`];
+  },
+
+  async getThirdParties(hostname) {
+    if (!hostname) {
+      return null;
```
jonathanKingston: This could be:
```
throw new Error("getThirdParties requires a valid hostname argument");
```

js/store.js
jonathanKingston: Yeah just make sure whatever `set` does `get` behaves the same way. Like I think some of the other code treats this as "set single item" where as `get` is "get all".

js/store.js
```diff
@@ -0,0 +1,81 @@
+const store = {
+  init() {
+    this.addListeners();
+  },
+
+  async getAll() {
+    return await browser.storage.local.get();
+  },
+
```
princiya: I will implement comments from this issue #18 in next PR

Issue #19, opened 2017-05-23
Testing platforms


jonathanKingston:
@princiya asked how we should be testing the extension so lets discuss that here.

This example has both functional and unit tests: https://github.com/Standard8/example-webextension

Unit tests likely make the most sense for the storage methods. For browser apis we potentially could mock them but I suspect it might make more sense to make those functional tests.

I don't think there needs to be a rush to decide right now which libraries are best.
Personally for unit testing I have used mocha and chai assert a lot. I also have used ava which seems very promise ready: https://github.com/avajs/ava

jonathanKingston:
@princiya the code https://github.com/electrolyfish/lightbeam-we/blob/master/js/store.js#L62 needs to be moved into non functional code, even if we make it require a console call to trigger it for now. However moving this to be a test would be neater.

PR #23, opened 2017-05-24
Basic test environment setup

princiya:
* As pointed out in this issue #19, the goal is to move [non-functional code](https://github.com/electrolyfish/lightbeam-we/blob/master/js/store.js#L62) into a test directory.
* Using this [example](https://github.com/Standard8/example-webextension) as a reference, I have setup a basic test environment which includes - Karma, Mocha and Chai.
* `get` methods from `js/store.js` are now testable.
* The store object is mocked and the corresponding methods are tested.

**To test this PR:**
`npm install`
`npm run test`

**Screenshot**

test-success-1


js/store.js
```diff
@@ -3,6 +3,10 @@ const store = {
     this.addListeners();
   },
 
+  getEmptyObject() {
```
princiya: added for testing purpose. once I have sinon-chrome setup working, this function will be removed.

js/store.js
```diff
@@ -59,6 +63,7 @@ const store = {
   }
 };
 
+/*
 store.getAll()
```
princiya: The goal of this PR is to move this code into the test directory

package.json
```diff
@@ -4,19 +4,43 @@
...
+    "lint": "npm-run-all lint:*",
+    "lint:eslint": "eslint --ext=.js,.json .",
+    "lint:webext": "web-ext -s add-on lint",
+    "test": "npm-run-all test:*",
+    "test:lint": "npm run lint",
+    "test:karma": "karma start"
   },
   "repository": {
     "type": "git",
-    "url": "git+https://github.com/pauljt/lightbeam-we.git"
```
princiya: haha... yeah ;)

princiya:
I am yet to add sinon-chrome part. maybe I do that in another PR? if so, this can be merged

groovecoder:
Nice start on the testing harness! `eslint` doesn't like a lot of stuff here, and I don't think `eslint` comes pre-configured to deal with `.json` files?

princiya:
@groovecoder there is an `eslint-plugin-json`, `eslint-plugin-mocha` to fix the existing lint errors. #TIL

jonathanKingston:
@princiya Lots of ESLint errors in Travis to fix:

https://travis-ci.org/electrolyfish/lightbeam-we/builds/236005297?utm_source=github_status&utm_medium=notification

Could you look into those?

princiya:
Yes, will do. Back to DO NOT MERGE :)

js/background.js
```diff
@@ -1,27 +1,26 @@
-'use strict';
-
-// When the user clicks browserAction icon in toolbar, run Lightbeam
-browser.browserAction.onClicked.addListener(runLightbeam);
-
-capture.init();
```
princiya: @biancadanforth are you able to invoke `capture.init()` from here?

biancadanforth: @princiya Yep. I've tested it in the browser.

PR #35, opened 2017-06-01
husky added. fixes #34

### LINKED ISSUE #34: Add git hook - `husky`
It is easy to forget to run linting or test tasks before committing or pushing code and this can result in a broken build.

A way to work around the above problem is to use git hooks and I suggest to use `husky`.

Open for discussions :)

princiya:
Issue #34 

Added precommit and prepush hooks


jonathanKingston:
LGTM merging this pretty sure @biancadanforth was happy with this. We can always pull this out if it becomes annoying.

PR #39, opened 2017-06-07
Store set method. Fixes #18

### LINKED ISSUE #18: Storage object: set method
In `browser.storage.local.set(keys)` method, the `keys` param is:
```
An object containing one or more key/value pairs to be stored in storage. If an item already exists, its value will be updated.
```
Our store structure looks like the following: #15 
```
websites: {
 'a.com': {},
 'b.com': {}
}
```
When we call `set()` to add/update hostnames, we need to pass the whole `websites` object. `websites` will be the key to the `browser.storage.set()`.
We need to update the local `websites` object with the right logic to add/update etc.

@jkt open for discussions, before I proceed :)

princiya:
Work in progress. Description to be updated shortly...

Issue #18 

Structure of the store is discussed here #15

### DIFF
--- a/js/store.js
+++ b/js/store.js
@@ -44,27 +44,23 @@ const store = {
     return null;
   },
 
-  async setFirstParty(hostname, data) {
+  setFirstParty(hostname, data) {
     if (!hostname) {
       throw new Error('setFirstParty requires a valid domain argument');
     }
 
-    const websites = await this._read();
+    const websites = clone(this._websites);
     websites[hostname] = data;
     this._write(websites);
   },
 
-  async setThirdParty(origin, target, data) {
+  setThirdParty(origin, target, data) {
     if (!origin) {
       throw new Error('setThirdParty requires a valid parent argument');
     }
 
-    const websites = await this._read();
-    const firstParty = websites[origin];
-
-    if (!firstParty) {
-      throw new Error('There is no firstParty matching the parent');
-    }
+    const websites = clone(this._websites);
+    const firstParty = clone(websites[origin]);
 
     if (!('thirdPartyRequests' in firstParty)) {
       firstParty['thirdPartyRequests'] = {};
--- a/js/store.js
+++ b/js/store.js
@@ -1,17 +1,18 @@
 const store = {
-  async init() {
-    const data = await this._read();
-    this._websites = data || null;
+  _websites: null,
+
+  init() {
+    this._read();
     this.addListeners();
   },
 
   async _read() {
-    return await browser.storage.local.get('websites');
+    this._websites = await browser.storage.local.get('websites');
+    return this._websites;
   },
 
   async _write(websites) {
     this._websites = websites;
-
     return await browser.storage.local.set({ websites });
   },
 
--- a/js/store.js
+++ b/js/store.js
@@ -41,24 +41,24 @@ const store = {
     return null;
   },
 
-  async setFirstParty(domain, data) {
-    if (!domain) {
+  async setFirstParty(hostname, data) {
+    if (!hostname) {
       throw new Error('setFirstParty requires a valid domain argument');
     }
 
     const websites = await this._read();
-    websites[domain] = data;
+    websites[hostname] = data;
 
     this._write(websites);
   },
 
-  async setThirdParty(parent, domain, data) {
-    if (!parent) {
+  async setThirdParty(origin, target, data) {
+    if (!origin) {
       throw new Error('setThirdParty requires a valid parent argument');
     }
 
     const websites = await this._read();
-    const firstParty = websites[parent];
+    const firstParty = websites[origin];
 
     if (!firstParty) {
       throw new Error('There is no firstParty matching the parent');
@@ -67,9 +67,9 @@ const store = {
     if (!('thirdPartyRequests' in firstParty)) {
       firstParty['thirdPartyRequests'] = {};
     }
-    firstParty['thirdPartyRequests'][domain] = data;
+    firstParty['thirdPartyRequests'][target] = data;
 
-    this.setFirstParty(parent, firstParty);
+    this.setFirstParty(origin, firstParty);
   },
 
   async remove() {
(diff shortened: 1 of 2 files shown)


js/store.js
```diff
@@ -29,31 +42,32 @@ const store = {
...
 
   async remove() {
     return await browser.storage.local.remove('websites');
   },
 
   addListeners() {
-    /*
-    * @todo update the code
-    */
+    browser.storage.onChanged.addListener((changes, area) => {
```
jonathanKingston: Unless we are expecting the storage to be written from elsewhere does this actually matter?

When write calls, this will update and might actually cause a place where several consecutive writes actually overwrite the previous data.

princiya: Yes, you are right. Removing this piece for now.

princiya: @jkt i think this listener is useful for the viz object to re-render the graph. what you think?

jonathanKingston:
@princiya is this ready for merge yet?

princiya:
nope. _websites is mutated and is wrongly cloned/copied. fixing this issue

js/store.js
```diff
@@ -1,19 +1,34 @@
...
+  _websites: null,
+
   init() {
-    this.addListeners();
+    this._read();
+  },
+
+  async _read() {
+    const data = await browser.storage.local.get('websites');
+    this._websites = clone(data.websites);
```
jonathanKingston: Why store it in `_websites` at all if you are going to make this block on a write?

Can you just read into `_websites` in `init` instead from the storage instead must making this read from the property?

princiya: I was a bit confused about to put this piece here or in `init`. So will move this into `init`..

princiya: ok.. thinking a bit and will update the code.

js/store.js
```diff
@@ -1,19 +1,34 @@
...
       throw new Error('getFirstParty requires a valid hostname argument');
     }
 
-    const storage = await this.getAll();
-    return storage.websites[hostname];
+    const websites = await this.getAll();
+    return websites[hostname];
   },
 
   async getThirdParties(hostname) {
```
biancadanforth: @princiya I am working on updating [capture.js](https://github.com/electrolyfish/lightbeam-we/blob/master/js/capture.js) per our discussion a few days ago (DNM PR coming shortly!). `hostname` is actually a key used for all three URLs (`document`, `target`, `origin`) by the webextension APIs I use in `capture.js`, so it may not be the best term when referring to the visited site's URL. What if we used `document` instead?

js/store.js
```diff
@@ -29,31 +44,39 @@ const store = {
...
+    websites[hostname] = data;
+    this._write(websites);
   },
 
-  async set(websites) {
-    /**
-      * @todo code to be updated in next PR
-    */
-    return await browser.storage.local.set({ websites });
+  setThirdParty(origin, target, data) {
```
biancadanforth: To piggy back on my first comment about using `document` instead of `hostname`, I believe here we are interested in the `document` URL, not `origin` URL. In extent, `setThirdParty(document, target, data)...`. I just created an issue to clarify the differences between `document`, `origin` and `target` URLs.

js/store.js
princiya: @biancadanforth @jkt IMO, we could use `hostname` here. this is a `get` query, and we do a get based on hostnames, irrespective of it coming from document, target or origin.

js/store.js
princiya: @biancadanforth @jonathanKingston here we could use `document` instead of `origin`, if the data comes from document and not origin.

princiya: or maybe, we leave it to `origin` and `target`? The idea with `setThirdParty` is it needs a parent or origin url, and the target url. It shouldn't worry about the url coming from document or origin based on the web extensions api. correct me please :)

js/store.js
```diff
@@ -1,59 +1,76 @@
...
+
+  async _write(websites) {
+    this._websites = clone(websites);
+
+    return await browser.storage.local.set({ websites });
   },
 
-  async getAll() {
-    return await browser.storage.local.get();
+  getAll() {
```
princiya: @jkt thanks for the suggestions. `browser.storage.local.get()` is now called only during `init`. I have removed `_read()` . Feedback?! :)

js/store.js
jonathanKingston: I agree with @princiya here in that origin isn't specific of which origin. If we wanted to be more specific we might change this to `firstPartyOrigin` and make the collection code use `document` to infer what the first party is... with the assumption that more avenues of data collection might be used later etc.
I'm fine to change `origin` to `firstPartyOrigin` in a follow up, perhaps also with a description of the arguments being that this will be part of this objects API.

biancadanforth: @princiya , @jonathanKingston : Thank you for the feedback! Capture.js is actually passing in `document` currently as the first argument for this method, not `origin`. I used `document` because I identify third party requests by comparing `document` to `target`.

What is meant by `origin` in this context? Is it the same as `origin` as defined in Issue #42 ?

Using `origin` here seems to be in contrast with why we're interested in `document` and `target` as explained in Issue #42 . What am I missing?

I agree that we could use JSDoc to help clarify these API questions as well as part of the scripts (mine included)!

js/store.js
biancadanforth: @princiya, @jonathanKingston I'm okay keeping `hostname`, since it's more general, though does it make sense to have a `getThirdParties` method for any hostname other than `document`?

js/store.js
princiya: @biancadanforth `origin` used in `setThirdParty(origin,...)` is just a formal parameter name and doesn't worry about the actual parameter value passed from capture.js. i.e. you are all good when you pass `document`. Hope this helps.

biancadanforth: Thanks @princiya That does help!

js/store.js
princiya: @biancadanforth `getThirdParties(hostname)` works uniformly irrespective of the actual value (hostname or document or any other website name) passed. `hostname` is just a formal argument :)

PR #43, opened 2017-06-11
Integrate storage API with data capture. Fixes #42.

### LINKED ISSUE #42: Integrate storage API with data capture
Update `capture.js` to call `store.setFirstParty` and `store.setThirdParty` methods.
Use consistent naming between arguments in `capture.js` and method parameters in `store.js`.

Questions to answer:
- What is the difference between `origin`, `target`, and `document` URLs?
- How often are `origin` URLs different from `document` URLs?
- What to do in these cases?

```javascript
// from capture.js  
async sendThirdParty(response) {
    const tab = await browser.tabs.get(response.tabId);
    const targetUrl = new URL(response.url);
    const originUrl = new URL(response.originUrl);

    if (targetUrl.hostname !== documentUrl.hostname) {
      const thirdPartyData = {
        document: documentUrl.hostname,
        target: targetUrl.hostname,
        origin: originUrl.hostname,
        requestTime: response.timeStamp
      };
      store.setThirdParty(documentUrl.hostname, targetUrl.hostname, thirdPartyData);
    }
  }
```

biancadanforth:
Now with calls to storage API instead of `console.log`!
I know I am still overriding some ESLint rules; but I believe they are a consequence of using separate scripts that are bundled. @jonathanKingston: When I tried to use modules (import/export) earlier for the web extension, Firefox didn't recognize the syntax. How can we avoid these manual overrides of ESLint rules?

### DIFF
--- a/js/capture.js
+++ b/js/capture.js
@@ -1,5 +1,9 @@
-/* eslint no-unused-vars: "off"*/
-/* eslint no-undef: "off" */
+/*
+* Listens for HTTP request responses, sending first- and
+* third-party requests to storage.
+*/
+
+capture.init();
 
 const capture = {
   init() {
--- a/js/capture.js
+++ b/js/capture.js
@@ -1,5 +1,5 @@
-/* eslint no-console: "off" */
-/* eslint no-unused-vars: "off" */
+/* eslint no-unused-vars: "off"*/
+/* eslint no-undef: "off" */
 
 const capture = {
   init() {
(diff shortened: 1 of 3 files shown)


js/capture.js
```diff
@@ -1,5 +1,5 @@
-/* eslint no-console: "off" */
-/* eslint no-unused-vars: "off" */
+/* eslint no-unused-vars: "off"*/
```
jonathanKingston: Hmm we should be careful with turning these things off, why is this needed?

jonathanKingston: My hunch is capture needs to be added as a global in the eslint file. We probably want a more explicit name eventually too that won't collide with other library globals.

jonathanKingston:
@biancadanforth ah sorry missed your comment here. I think we can add `capture` as a global which should prevent the overrides.

I think you may need to prevent the unused use of `capture` still which something like:
```
// eslint-disable-next-line no-unused-vars
```

This is way more explicit to prevent this globally disabling everything.

jonathanKingston:
Moving: https://github.com/electrolyfish/lightbeam-we/blob/master/js/background.js#L7 to this file removes the need for this unused vars anyway.

I'm not sure how store is not having the globals issue though.

biancadanforth:
@jonathanKingston  `store` isn't having the same unused variables issue because it isn't actually added yet to the `manifest.json` file like `capture` is.

I will move `capture.init();` into `capture.js`; that's an excellent idea.

I will hold off adding global variables to `eslint` for now, since it is not necessary currently.

jonathanKingston:
> store isn't having the same unused variables issue because it isn't actually added yet to the manifest.json file like capture is.

The manifest doesn't impact which files to lint.

I can't think of any alternatives right now. Explicit by being longer so they don't collide with other globals elsewhere. More like:  `captureData`, `Lightbeam.capture`...

js/capture.js
```diff
@@ -1,5 +1,9 @@
-/* eslint no-console: "off" */
-/* eslint no-unused-vars: "off" */
+/*
+* Listens for HTTP request responses, sending first- and
+* third-party requests to storage.
+*/
+
+capture.init();
```
jonathanKingston: Despite the fact JS has hoisting, I prefer this after it's definition. However I'm gonna merge this.

PR #47, opened 2017-06-13
Top level grid layout for UI.

biancadanforth:
Just saving progress thus far. Experimenting with CSS Grid.


js/background.js
```diff
@@ -3,6 +3,9 @@
 // When the user clicks browserAction icon in toolbar, run Lightbeam
 browser.browserAction.onClicked.addListener(runLightbeam);
 
+// temporarily saving myself a step of clicking the icon
+runLightbeam();
```
jonathanKingston: Remove this yeah?

index.html
```diff
@@ -9,12 +9,134 @@
...
+                Cookies
+              </button>
+            </div>
+          </div>
+          <div>
+            <h2>Filter</h2>
+            <h2 class="hide">Hide</h2>
+            <div class="graph-filter">
+              <button class="footer-button button--filter active">Recent Site</button>
+              <button class="button--filter">Last 10 Sites</button>
```
jonathanKingston: I'm not sure why `--filter` is needed here over just:

```
.graph-filter > button {
```

index.html
```diff
@@ -9,12 +9,134 @@
...
+                </svg>
+                Cookies
+              </button>
+            </div>
+          </div>
+          <div>
+            <h2>Filter</h2>
+            <h2 class="hide">Hide</h2>
+            <div class="graph-filter">
+              <button class="footer-button button--filter active">Recent Site</button>
```
jonathanKingston: How come this is a `footer-button` also?

css/style.css
```diff
@@ -1,7 +1,355 @@
...
+
+/* ----------- UI - Graph Filter ----------- */
+
+footer div:last-child {
+  position: relative;
+  grid-column-start: 5;
+}
+
+.hide {
+  position: absolute;
```
jonathanKingston: Why the need for position absolute here? It looks like an antipattern for this.

css/style.css
```diff
@@ -1,7 +1,355 @@
...
+.footer-button.active {
+  background-color: #73A4B8;
+  border-top: 2px solid #6FC3E5;
+}
+
+.footer-button.active:hover {
+  border-top-color: #fff;
+}
+
+.col-1 {
```
jonathanKingston: This seems like a copy of a grid based layout using classes. I'm not sure using these row and col classes is a good pattern here. Maybe we should discuss other solutions (right now it's late so not sure).

css/style.css
```diff
@@ -1,7 +1,355 @@
...
+.logo {
+  max-width: 100%;
+}
+
+aside h2 {
+  margin-top: 20px;
+  color: #EAEAEA;
+}
+
+.nav__button {
```
jonathanKingston: We have mixed use of BEM and non BEM it seems here which is adding code bloat from what I can see.

For some of the code modifiers like `.active` are being used elsewhere `--filter` for modifiers.

For scoping there is `.nav__button` but also `.graph h2`.

I'm really not a fan of BEM which is a coding style choice which perhaps should be discussed but certainly mixing the two for what seems like similar use cases should be avoided.

For the complexity of this CSS it's likely BEM can easily be avoided and more semantic tags could be used like:
```
nav > button
```

index.html
```diff
@@ -9,12 +9,134 @@
...
-    <canvas id="canvas" width="300" height="300"></canvas>
+    <aside class="controls">
+    <header>
+      <img class="logo" role="banner" src="images/lightbeam_150x45.png
+">
+    </header>
+    <nav role="navigation">
+      <h2>Visualization</h2>
+      <div class="nav__vis">
+        <button class="nav__button nav__button--vis"><img class="icon icon--button" src="images/lightbeam_icon_graph.png">Graph</button>
```
jonathanKingston: `<button class="icon graph-icon">Graph</button>`

 looks like it could be used over the other code:
```
<button class="nav__button nav__button--vis"><img class="icon icon--button" src="images/lightbeam_icon_graph.png">Graph</button>
```

jonathanKingston: As there are limited icons, hard coding those to classes simplifies the css and html.

jonathanKingston: It also looks like it solves the translate code too. Look into using `::before` here and using flex on the button itself to centre align them.

css/style.css
```diff
@@ -1,7 +1,355 @@
...
+}
+
+h3 {
+  font-size: 1.2em;
+  color: #73A4B8;
+  text-transform: uppercase;
+}
+
+button {
+  position: relative;
```
jonathanKingston: Using relative and absolute here? Is this for animations? Absolute really should be avoided at most costs.

index.html
```diff
@@ -9,12 +9,134 @@
...
+        <a href="https://support.mozilla.org/en-US/kb/disable-or-remove-add-ons">Uninstall Lightbeam</a>
+      </div>
+    </nav>
+    </aside>
+    <main role="main">
+      <section class="top-bar">
+        <div class="top-bar__summary">
+          <span>
+            <h2>Data Gathered Since</h2>
+            <h3 id="date-gathered">June 13, 2017</h3>
```
jonathanKingston: Can you add TODO HTML comments to anything hard coded? It's easy to miss things later on and forget it's not actually working.

index.html
biancadanforth: Actually just removing positioning and using `display:flex; align-items: center;` worked like a charm, no pseudo-element needed!

index.html
biancadanforth: Done!

index.html
biancadanforth: That's how I was making it look "active" before I added the active class and then forgot to remove it. :)

index.html
biancadanforth: I have greatly simplified the button classes. Now there is just `active` and `double-icon`.

index.html
jonathanKingston: The pseudo element is for the image to replace the image element and use classes to specify the icons in the CSS.

css/style.css
biancadanforth: Replaced all instances of absolute with either flexbox or grid.

css/style.css
biancadanforth: Where possible, I now use type selectors and more child selectors. The BEM class naming convention has been dropped.

css/style.css
biancadanforth: Removed!

index.html
biancadanforth: Okay; thanks for clarifying here and in slack. `::before` pseudoelement in place!

css/style.css
biancadanforth: I removed these classes; I've largely kept the styles the same (using grid properties) however with more semantic class names; open to more discussion.

index.html
```diff
@@ -9,12 +9,130 @@
...
     <script src="js/store.js" type="text/javascript"></script>
     <script src="js/viz.js" type="text/javascript"></script>
   </head>
   <body>
-    <h1>Lightbeam</h1>
-    <canvas id="canvas" width="300" height="300"></canvas>
+    <aside>
+    <header>
+      <img class="logo" role="banner" src="images/lightbeam_150x45.png
+">
```
jonathanKingston: Nit: newline.

css/style.css
```diff
@@ -1,7 +1,388 @@
...
+}
+
+.blocked-sites {
+  grid-column-start: 2;
+  grid-row-start: 2;
+}
+
+.blocked-sites::before {
+  background-image: url('data:image/svg+xml,<svg style="fill: rgb(224, 42, 97)" version="1.1" xmlns="http://www.w3.org/2000/svg"><circle cx="8" cy="8" r="6"></circle></svg>'), url('data:image/svg+xml,<svg style="margin-left: 15px; fill: rgb(224, 42, 97)" version="1.1" xmlns="http://www.w3.org/2000/svg"><polygon points="0,14 7,2 14,14"></polygon></svg>');
+  width: 30px;
```
jonathanKingston: Nit: What is distinct about `.blocked-sites` and `.watched-sites` that require `width: 30px;` that `.connections` doesn't? Can they share a selector to do this?

```
.bargh > button::before {
  width: 30px;
}
```

I haven't looked at the layout so not sure, however if the icon itself shouldn't really be tied to width, margin etc.

css/style.css
```diff
@@ -1,7 +1,388 @@
...
+}
+
+.about {
+  border-radius: 3px 0 0 3px;
+  margin-top: 5px;
+}
+
+.about::before {
+  background-image: url('../images/lightbeam_icon_about.png');
+  margin-right: 0;
```
jonathanKingston: `margin: 0;` same as my comment about the other `::before` can this use a different selector so we don't need multiple `margin-right: 0;`. This is super nit picky but just wanted to ask :)

PR #60, opened 2017-06-16
Nit follow ups for layout. Fixes #55.

### LINKED ISSUE #55: Nit follow ups for layout
As mentioned in: https://github.com/electrolyfish/lightbeam-we/pull/47/files there were some comments that were mostly questions.

**Comment 1:**
```css
.about::before {
  background-image: url('../images/lightbeam_icon_about.png');
  margin-right: 0;
```
> `margin: 0;` same as my comment about the other `::before` can this use a different selector so we don't need multiple `margin-right: 0;`. This is super nit picky but just wanted to ask :)

**Comment 2:**
```css
.blocked-sites::before {
  background-image: url('data:image/svg+xml,<svg style="fill: rgb(224, 42, 97)" version="1.1" xmlns="http://www.w3.org/2000/svg"><circle cx="8" cy="8" r="6"></circle></svg>'), url('data:image/svg+xml,<svg style="margin-left: 15px; fill: rgb(224, 42, 97)" version="1.1" xmlns="http://www.w3.org/2000/svg"><polygon points="0,14 7,2 14,14"></polygon></svg>');
  width: 30px;
```

> Nit: What is distinct about `.blocked-sites` and `.watched-sites` that require `width: 30px;` that `.connections` doesn't? Can they share a selector to do this?
```
.bargh > button::before {
  width: 30px;
}
```
> I haven't looked at the layout so not sure, however if the icon itself shouldn't really be tied to width, margin etc.

**Comment 3:**
```html
      <img class="logo" role="banner" src="images/lightbeam_150x45.png
">
```

> Nit: unnecessary newline.

biancadanforth:
Comments from Issue #55:
Comment 1: Fixed with a new `info-panel` button class.
Comment 2: `.blocked-sites` and `.watched-sites` both have two SVG icons instead of one (unlike any of the other controls buttons. Yes they can share a selector! Added a `two-icons` button class for that.
I use the width to specify the width of the `::before` pseudoelement. In the case of two icons, that space, originally specified in `button::before` for general icon buttons, needs to be increased from 15px to 30px for this case of two icons (each of width 15px). Alternative ideas on this welcome!
Comment 3: Typo fixed!

### DIFF
--- a/index.html
+++ b/index.html
@@ -16,8 +16,7 @@
   <body>
     <aside>
     <header>
-      <img class="logo" role="banner" src="images/lightbeam_150x45.png
-">
+      <img class="logo" role="banner" src="images/lightbeam_150x45.png">
     </header>
     <nav role="navigation">
       <h2>Visualization</h2>
@@ -83,15 +82,15 @@
             <h2><span id="view">Graph</span> View</h2>
             <!-- @todo display list or graph view, update view based on filter and controls -->
             <!-- @todo remove the hard-coded width, height values -->
-            <canvas id="canvas" width="900" height="300"></canvas>
+            <canvas id="canvas" width="100%" height="100%"></canvas>
           </div>
           <div class="info-panel-controls">
             <!-- @todo check purpose of website icon -->
-            <button class="website"></button>
+            <button class="info-panel website"></button>
             <!-- @todo add help sidebar -->
-            <button class="help"></button>
+            <button class="info-panel help"></button>
             <!-- @todo add about sidebar -->
-            <button class="about"></button>
+            <button class="info-panel about"></button>
           </div>
         </div>
         <footer>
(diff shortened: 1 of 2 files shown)


index.html
```diff
@@ -83,15 +82,15 @@ <h1 id="filter">Recent Site</h1>
             <h2><span id="view">Graph</span> View</h2>
             <!-- @todo display list or graph view, update view based on filter and controls -->
             <!-- @todo remove the hard-coded width, height values -->
-            <canvas id="canvas" width="900" height="300"></canvas>
+            <canvas id="canvas" width="100%" height="100%"></canvas>
```
biancadanforth: @princiya I updated the canvas size to make it fit inside its parent element. I hope that's all right with you!

PR #61, opened 2017-06-16
Add structural/layout fixes to top bar. Fixes #56.

### LINKED ISSUE #56: Header fixes for small screens
This isn't about implementing a full responsive layout that can come later.

However on smaller screens some of the text wrapping causes layout issues:

selection_692

The HTML for this top bar is also:
```
<section class="top-bar">
        <div>
          <span>
            <h2>Data Gathered Since</h2>
            <!-- @todo get date-start from store -->
            <h3 id="date-start">June 13, 2017</h3>
          </span>
          <span>
            <h2>You Have Visited</h2>
            <!-- @todo get num-sites-visited from store -->
            <h3><span id="num-sites-visited">6</span> Sites</h3>
          </span>
          <span>
            <h2>You Have Connected With</h2>
            <!-- @todo get num-third-parties from store -->
            <h3><span id="num-third-parties">16</span> Third Party Sites</h3>
          </span>
        </div>
        <!-- @todo add tracking protection and toggle functionality -->
        <div class="tracking-protection">
          <h2>Tracking Protection</h2>
          <div class="toggle-outer">
            <div class="toggle-inner">
            </div>
            <!-- @todo toggle-label to reflect state of tracking protection -->
            <div class="toggle-label">
            Off
            </div>
          </div>
        </div>
      </section>
```

I'm fairly convinced with grid this could be:
```
<section id="top-bar">
  <dl class="info-headings">
    <dt>
       Data Gathered Since
    </dt>
    <dd id="date-start">
      <time datetime="2017-06-13T11:00">June 13, 2017</time>
    </dd>

    <dt>
       You Have Visited
    </dt>
    <dd>
       <var id="num-sites-visited">6</var> Sites
    </dd>

    <dt>
      You Have Connected With
    </dt>
    <dd>
       <var id="num-third-parties">16</var> Third Party Sites
     </dd>
  </dl>

  <div class="tracking-protection">
    <label for="tracking-protection-control">Tracking Protection</label>
    <div class="toggle-switch"><input type="checkbox" id="tracking-protection-control" /></div>
  </div>
</section>
```

Comments:

- Using span's for block like items is usually a bad sign as they are an inline element by default so using div's is slightly better
- Now we have grid using dl's is certainly much easier for this use-case
- Using a real input element and using before means we get some of the accessibility goodness from the browser
- Using grid here should let it expand downwards and also later be reorganisable for mobile.
- I used var and time for a little bit of semantic goodness
- I can also see that flex could also be a good fit here, standards people just made it ok to wrap `<dd>` and `<dt>` with a `<div>` whatever works :)

biancadanforth:
- Changed structure of `.top-bar` to use more semantic elements than `<span>`.
- Tracking protection toggle now uses `<input>` and `<label>` elements. Note: still using `<span>` elements for custom appearance. Added a @todo to couple checkbox to custom `<span>` element.
- Added a `min-width` property to `<main>`, so the `.top-bar` text (or content elsewhere in `<main>`) doesn't break. I plan to make a follow-up issue to make UI responsive.

### DIFF
--- a/css/style.css
+++ b/css/style.css
@@ -35,7 +35,7 @@ body {
   font-weight: normal; 
 }
 
-h1, h2, h3 {
+h1, h2, h3, dt, dd, label {
   margin: 0;
   font-weight: 400;
 }
@@ -44,16 +44,20 @@ h1 {
   font-size: 1.75em;
 }
 
-h2 {
+h2, dt, label {
   font-size: .75em;
   font-weight: 700;
   text-transform: uppercase;
 }
 
-h3 {
+h3, dd {
   font-size: 1.2em;
   color: var(--secondary-text-color);
   text-transform: uppercase;
+}
+
+var {
+  font-style: normal;
 }
 
 button {
@@ -183,6 +187,7 @@ aside h2 {
 /* ----------- UI - Top Bar ----------- */
 
 main {
+  min-width: 780px;
   display: grid;
   grid-template-rows: 10% 90%;
   background-color: var(--primary-color);
@@ -196,11 +201,11 @@ main {
   padding: 20px;
 }
 
-.top-bar > div {
+.top-bar > dl {
   display: flex;
 }
 
-.top-bar > div > span {
+.top-bar > dl > div {
   margin-right: 15px;
 }
 
@@ -209,7 +214,7 @@ main {
   align-items: center;
 }
 
-.toggle-outer {
+.toggle-switch {
   display: flex;
   align-items: center;
   justify-content: space-around;
@@ -220,14 +225,19 @@ main {
   border-radius: 2px;
 }
 
-.toggle-inner {
+input[type="checkbox"] {
+  -moz-appearance: none;
+  margin: 0;
+}
+
+.toggle {
   height: 20px;
   width: 30px;
   background-color: var(--primary-color);
   border-radius: 5px;
 }
 
-.toggle-label {
+.toggle-status {
   width: 40%;
   font-size: .75em;
   font-weight: 700;
--- a/index.html
+++ b/index.html
@@ -44,32 +44,38 @@
     </aside>
     <main role="main">
       <section class="top-bar">
-        <div>
-          <span>
-            <h2>Data Gathered Since</h2>
-            <!-- @todo get date-start from store -->
-            <h3 id="date-start">June 13, 2017</h3>
-          </span>
-          <span>
-            <h2>You Have Visited</h2>
+        <dl class="info-headings">
+          <div>
+            <dt>Data Gathered Since</dt>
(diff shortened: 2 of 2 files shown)


css/style.css
```diff
@@ -220,14 +225,18 @@ main {
   border-radius: 2px;
 }
 
-.toggle-inner {
+input[type="checkbox"] {
+  display: none;
```
jonathanKingston: Display none will remove all the accessibility benefits of this, use `opacity: 0;`

biancadanforth: My understanding of `display:none` versus `opacity:0` (though I can't find an official resource that states unequivocally):
- `display: none` means that the element in question will not appear on the page at all (although you can still interact with it through the DOM). When I inspect the HTML for the page, the `<input>` element is present, though flagged as having a `display:none`. Doesn't this mean you still get the accessibility benefits? Also importantly: there will be no space allocated for it between the other elements.
- `opacity: 0` means that unlike `display: none`, space is allocated for it on the page. I don't want space to be allocated for it on the page, so wouldn't I want `display:none` here? If I use opacity, then I'd have to add a number of other styles to prevent the checkbox from taking up physical space on the page:
```css
input[type="checkbox"] {
  /*display: none;*/
  opacity: 0;
  height: 0;
  width: 0;
  margin: 0;
  border: 0;
}
```
And for some reason, the element still has a border around it; even though inspect tells me that the `border:0` rule is being enforced.

jonathanKingston: > Doesn't this mean you still get the accessibility benefits?

No tabbing and space to select won't work.

> Also importantly: there will be no space allocated for it between the other elements.

I'm not sure what you mean, I assume you mean it will be bad in the layout. Firefox just fixed allowing `-moz-appearance: none; height: 0;..` I think, if not there are other ways like absolute positioning (valid use case if other methods of removing from the layout doesn't work).

>  prevent the checkbox from taking up physical space on the page

Well `margin` `border` `padding` should likely be applied to all elements anyway.
A few attributes aren't really an issue, if we need the same styles again we can always create another class like `.visually-hidden`.

`visibility: hidden;` also makes the checkbox non interactive too so that is out.

> the element still has a border around it; even though inspect tells me that the border:0 rule is being enforced.

Yeah check `appearance: none;` if not hide it off page with indent or positioning. Form fields like a few other elements have special appearance by default which makes these styles persist. To my knowledge they literally just fixed the height and width resetting too which should hopefully help.

jonathanKingston: Also a few properties is way simpler than the tonne of JS to handle keyboard tab, space and clicking to change the class on a div.

biancadanforth: Okay I am understanding you now. Thank you for the reply! I have made the changes.

PR #62, opened 2017-06-16
Remove moz-extension:* paths from capture. Fixes #49.

### LINKED ISSUE #49: MozExtension paths should be excluded from storage
In capture we exclude `about:` we should also include `moz-extension:` to be ignored.

biancadanforth:
Added a helper function called `shouldStore` to filter out unwanted first- and third-party requests, rather than just first-party only.

Also removed a duplicate `js/store.js` in the `manifest.json` file.

### DIFF
--- a/js/capture.js
+++ b/js/capture.js
@@ -10,10 +10,25 @@ const capture = {
   addListeners() {
     // listen for each HTTP response
     browser.webRequest.onResponseStarted.addListener(
-      this.sendThirdParty,
+      (response) => this.sendThirdParty(response),
       {urls: ['<all_urls>']});
     // listen for tab updates
-    browser.tabs.onUpdated.addListener(this.sendFirstParty);
+    browser.tabs.onUpdated.addListener(
+      (tabId, changeInfo, tab) => {
+        this.sendFirstParty(tabId, changeInfo, tab);
+      });
+  },
+
+  // Returns true if the request should be stored, otherwise false.
+  shouldStore(tab) {
+    const documentUrl = new URL(tab.url);
+    // ignore about:*, moz-extension:* & non-visible tabs (like dev tools)
+    if (documentUrl.protocol !== 'about:'
+      && documentUrl.protocol !== 'moz-extension:'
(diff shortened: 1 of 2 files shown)


js/capture.js
```diff
@@ -10,10 +10,22 @@ const capture = {
   addListeners() {
     // listen for each HTTP response
     browser.webRequest.onResponseStarted.addListener(
-      this.sendThirdParty,
+      this.sendThirdParty.bind(this),
```
jonathanKingston: I prefer:
```
(tabId, changeInfo, tab) => {
  this.sendThirdParty(tabId, changeInfo, tab);
}
```
or
```
(...args) => {
  this.sendThirdParty(...args);
}
```

over:
```
this.sendThirdParty.bind(this)
```

As you aren't creating a new function, there aren't really perf benefits I don't think but it just feels more explicit.

jonathanKingston: Feel free to ignore though

biancadanforth: I like the arrow functions better too. Lexical scope makes sense, and you can see the arguments passed in too. Changed!

js/capture.js
```diff
@@ -10,10 +10,25 @@ const capture = {
   addListeners() {
     // listen for each HTTP response
     browser.webRequest.onResponseStarted.addListener(
-      this.sendThirdParty,
+      (response) => this.sendThirdParty(response),
```
jonathanKingston: Please add a follow up bug to make this use braces. Having the implied return may in future cause issues as this method returns a promise...blocking requests for example would delay.

biancadanforth: Are you saying to update it like this?:
```javascript  
 addListeners() {
    // listen for each HTTP response
    browser.webRequest.onResponseStarted.addListener(
      (response) => {
        this.sendThirdParty(response);
      },
      {urls: ['<all_urls>']});
    // ...
```
Which method returns a promise?

You are concerned that without the curly braces, the argument may not be read correctly by the webRequest API as a callback function? Could you explain a little more.

PR #63, opened 2017-06-18
D3 visualization

biancadanforth:
Initial D3 visualization draws first and third party requests to the page.


js/lightbeam.js
```diff
@@ -1,9 +1,53 @@
...
+  // set up area of the page to put the graph
+  const svg = d3.select('svg');
+  const width = svg.attr('width');
+  const height = svg.attr('height');
+
+  // set up simulation
+  const simulation = d3.forceSimulation().nodes(nodes_data);
+
+  // add forces
+  simulation
```
jonathanKingston: Why is this not in viz?

js/lightbeam.js
```diff
@@ -1,9 +1,53 @@
...
+      // third parties likely don't have favicons
+      node.favicon = '';
+      node.party = 'third';
+      nodes_data.push(node);
+      links_data.push(link);
+    }
+  }
+
+  // set up area of the page to put the graph
+  const svg = d3.select('svg');
```
jonathanKingston: This should be in viz no?

.eslintignore
```diff
@@ -0,0 +1,2 @@
+# Ignore external libraries
+js/lib/*
```
jonathanKingston: Use npm instead?

.gitignore
```diff
@@ -1,3 +1,4 @@
 .DS_Store
 node_modules
 npm-debug.log
+js/lib/*
```
jonathanKingston: If moving to npm remove this.

js/lightbeam.js
```diff
@@ -1,9 +1,53 @@
 async function renderGraph() {
-  const canvas = document.getElementById('canvas');
-  const context = canvas.getContext('2d');
   const websites = await store.getAll();
 
-  viz.draw(context, websites);
+  // reformat data for D3
```
jonathanKingston: This looks like it is coupled to the visualisations. It shouldn't be.

This code might:
- Change hostname paths and clean replicated data
- Query first party and third parties in a an order it chooses

It shouldn't be changing output to match how a rendering might display it.

js/lightbeam.js
```diff
@@ -1,9 +1,53 @@
...
+    for (const thirdParty in websites[firstParty].thirdPartyRequests) {
+      const node = {};
+      const link = {};
+      link.source = firstParty;
+      link.target = thirdParty;
+      node.hostname = thirdParty;
+      // third parties likely don't have favicons
+      node.favicon = '';
+      node.party = 'third';
+      nodes_data.push(node);
```
jonathanKingston: I thought we were using camel case?

js/lightbeam.js
biancadanforth: What are you proposing I do here instead?

js/lightbeam.js
biancadanforth: Moved

js/lightbeam.js
biancadanforth: Moved

js/lightbeam.js
biancadanforth: Yes, we are. This was just based on a tutorial I was using at the time. Changed!

.eslintignore
biancadanforth: If I use npm, how do I include the d3 library script in the viz.js file? All solutions so far use `require` or `import` (would need to transpile).

PR #67, opened 2017-06-19
Initial d3 visualization

biancadanforth:
Updates with real data, but updates on a timer, rather than being notified that something has changed from storage.
@todo:
* Consider using canvas instead of svg elements
* Consider using a callback in storage to update visualization
* Currently the storage data is reformatted with a function transformData; consider alternatives.

### DIFF
--- a/css/style.css
+++ b/css/style.css
@@ -1,11 +1,13 @@
 /* ----------- VISUALIZATION ----------- */
 
-.circle {
-  display: inline-block;
-  width: 100px;
-  height: 100px;
-  border-radius: 50px;
-  background-color: red;
+.links line {
+  stroke: #999;
+  stroke-opacity: 0.6;
+}
+       
+.nodes circle {
+  stroke: #fff;
+  stroke-width: 1.5px;
 }
 
 /* ----------- UI - General ----------- */
@@ -258,6 +260,13 @@ input[type="checkbox"] {
   display: flex;
   align-items: flex-start;
   justify-content: space-between;
+  height: 100%;
+}
+
+.vis > div > div {
+  display: flex;
+  flex-direction: column;
+  height: 100%;
 }
 
 .vis h1 {
--- a/index.html
+++ b/index.html
@@ -11,7 +11,9 @@
     <link rel="stylesheet" href="css/style.css" type="text/css">
     <link href="https://fonts.googleapis.com/css?family=Open+Sans" rel="stylesheet">
     <script src="js/storeChild.js" type="text/javascript"></script>
+    <script src="node_modules/d3/build/d3.min.js"></script>
     <script src="js/viz.js" type="text/javascript"></script>
+    <script src="js/lightbeam.js" type="text/javascript"></script>
   </head>
   <body>
     <aside>
@@ -88,7 +90,7 @@
             <h2><span id="view">Graph</span> View</h2>
             <!-- @todo display list or graph view, update view based on filter and controls -->
             <!-- @todo remove the hard-coded width, height values -->
-            <canvas id="canvas" width="100%" height="100%"></canvas>
+            <svg id="visualization" width="100%" height="100%"></svg>
           </div>
           <div class="info-panel-controls">
             <!-- @todo check purpose of website icon -->
@@ -139,6 +141,5 @@
         </footer>
       </section>
     </main>
-    <script src="js/lightbeam.js" type="text/javascript"></script>
   </body>
 </html>
--- a/js/lightbeam.js
+++ b/js/lightbeam.js
@@ -1,9 +1,52 @@
-async function renderGraph() {
-  const canvas = document.getElementById('canvas');
-  const context = canvas.getContext('2d');
-  const websites = await storeChild.getAll();
+// transforms the object of nested objects 'websites' into a
+// usable format for d3
+function transformData(websites) {
+  const nodes = [];
+  const links = [];
+  const firstParties = new Set();
+  const thirdParties = new Set();
+  // make first party nodes
+  for (const firstParty in websites) {
+    const firstPartyNode = createWebsite(firstParty);
+    firstParties.add(firstPartyNode);
+    nodes.push(firstPartyNode);
+    // make third party nodes and links
+    if (websites[firstParty].thirdPartyRequests) {
+      for (const thirdParty in websites[firstParty].thirdPartyRequests) {
+        const thirdPartyNode = createWebsite(thirdParty);
+        thirdParties.add(thirdPartyNode);
+        if (!firstParties.has(thirdPartyNode)) {
+          nodes.push(thirdPartyNode);
+        }
+        const link = {};
+        link.source = firstParty;
+        link.target = thirdParty;
+        links.push(link);
+      }
 
-  viz.draw(context, websites);
+    }
+  }
+
+  function createWebsite(website) {
+    const websiteOutput = {
+      hostname: website,
+      favicon: websites[website] ? websites[website].faviconUrl : '',
+      party: websites[website] ? 'first' : 'third'
+    };
+    return websiteOutput;
+  }
+
+  return {
+    nodes: nodes,
+    links: links
+  };
 }
 
-renderGraph();
+async function renderGraph() {
+  const websites = await storeChild.getAll();
+  const transformedData = transformData(websites);
+  viz.init(transformedData.nodes, transformedData.links);
+}
+
+window.onload = renderGraph;
+
--- a/js/viz.js
+++ b/js/viz.js
@@ -1,29 +1,101 @@
 // eslint-disable-next-line no-unused-vars
 const viz = {
-  draw(context, websites) {
-    const pi = Math.PI * 2;
-    let x = 50, y = 50;
+  init(nodes, links) {
+    const svg = d3.select('svg');
+    const nodesGroup = svg.append('g');
+    nodesGroup.attr('class', 'nodes');
+    const linksGroup = svg.append('g');
+    linksGroup.attr('class', 'links');
+    this.allCircles = nodesGroup.selectAll('.node');
+    this.allLines = linksGroup.selectAll('.link');
 
-    context.fillStyle = 'white';
-    context.beginPath();
+    // D3 needs explicit pixel values for 'center_force'
+    const visualization = document.getElementById('visualization');
+    const width = visualization.getBoundingClientRect().width;
+    const height = visualization.getBoundingClientRect().height;
 
-    for (const website in websites) {
-      context.moveTo(x, y);
-      context.arc(x, y, 10, 0, pi);
-      context.fillText(website, x, y);
+    this.simulation = d3.forceSimulation(nodes);
+    this.simulation.force('charge', d3.forceManyBody());
+    const linkForce = d3.forceLink(links);
+    // link source/target values are uniquely identified by
+    // the hostname property
+    linkForce.id(function (d) {
+      return d.hostname;
+    });
+    linkForce.distance(50);
+    this.simulation.force('link', linkForce);
+    this.simulation.force('center', d3.forceCenter(width/2, height/2));
+    this.simulation.alphaTarget(1);
+    this.simulation.on('tick', () => {
+      this.ticked();
+    });
 
-      let x1 = x;
-      for (const thirdParty in websites[website].thirdPartyRequests) {
-        x1+=150;
-        context.moveTo(x1, y);
-        context.arc(x1, y, 5, 0, pi);
-        context.fillText(thirdParty, x1, y);
-      }
+    // initial render
+    this.update(nodes, links);
+  },
 
-      x+=20; y+=20;
-    }
+  ticked() {
 
-    context.stroke();
-    context.fill();
+    this.allCircles
+      .attr('cx', function(d) {
+        return d.x;
+      })
+      .attr('cy', function(d) {
+        return d.y;
+      });
+
+    this.allLines
+      .attr('x1', function(d) {
+        return d.source.x;
+      })
+      .attr('y1', function(d) {
+        return d.source.y;
+      })
+      .attr('x2', function(d) {
+        return d.target.x;
+      })
+      .attr('y2', function(d) {
+        return d.target.y;
+      });
+  },
+
+  update(nodes, links) {
+    // determine which nodes to keep, remove and add: the update selection
+    this.allCircles = this.allCircles.data(nodes, function(d) {
+      return d.hostname;
+    });
+
+    // remove old nodes: the exit selection
+    const oldNodes = this.allCircles.exit();
+    oldNodes.remove();
+
+    // add new nodes: the enter selection
+    let newNodes = this.allCircles.enter();
+    newNodes = newNodes.append('circle');
+    newNodes.attr('fill', 'red');
+    newNodes.attr('r', 5);
+    this.allCircles = newNodes.merge(this.allCircles);
+
+    // determine which links to keep, remove and add, the update selection
+    this.allLines = this.allLines
+      .data(links, function(d) {
+        return `${d.source.hostname}-${d.target.hostname}`;
+      });
+
+    // remove old links
+    const oldLinks = this.allLines.exit();
+    oldLinks.remove();
+
+    // add new links
+    let newLinks = this.allLines.enter();
+    newLinks = newLinks.append('line');
+    this.allLines = newLinks.merge(this.allLines);
+
+    // Update and restart the simulation.
+    this.simulation.nodes(nodes);
+    const linkForce = this.simulation.force('link');
+    linkForce.links(links);
+    this.simulation = this.simulation.alpha(1);
+    this.simulation.restart();
   }
 };
--- a/package.json
+++ b/package.json
@@ -23,7 +23,9 @@
     "url": "https://github.com/electrolyfish/lightbeam-we/issues"
   },
   "homepage": "https://github.com/electrolyfish/lightbeam-we#readme",
-  "dependencies": {},
+  "dependencies": {
+    "d3": "4.9.1"
+  },
   "devDependencies": {
     "chai": "^3.5.0",
     "eslint": "^3.19.0",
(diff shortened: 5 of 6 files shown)


js/lightbeam.js
```diff
@@ -1,9 +1,33 @@
 async function renderGraph() {
-  const canvas = document.getElementById('canvas');
-  const context = canvas.getContext('2d');
   const websites = await store.getAll();
 
-  viz.draw(context, websites);
+  // reformat data for D3
```
biancadanforth: **jonathan Kingston:**
This looks like it is coupled to the visualisations. It shouldn't be.
This code might:
    - Change hostname paths and clean replicated data
    - Query first party and third parties in a an order it chooses
It shouldn't be changing output to match how a rendering might display it.

**biancadanforth**:
What are you proposing I do here instead? We don't want data from the store to worry about which visualization library is used, but the data needs to be of a certain format for drawing in D3 (i.e. it expects an array of objects, not an object of nested objects).
I agree with you that the order in which it assembles `nodes_data` and `links_data` could be different from the order those keys appear in `websites`, but why does that matter?

**jonathanKingston**:
By ordering I meant async nature. I could imagine this code querying the data in such a way that it draws in order of last used or most common or some other factor.
I'm now thinking the structure passed to vis might be better expressed as individual websites.
If you are manipulating data to match the nodes being drawn it likely best belongs where it is about to be drawn.
Now that I have thought about the code a little more I thin there are a few avenues:
1. Vis code could accept big blobs and reshuffle itself, however that binds the vis code to the nested structure of how he storage currently works (Long term I think this is a little cumbersome so likely we shouldn't do this)
2. This code could call individual methods on viz like viz.drawSite(hostname, {icon: "..", other: things}) and viz.connectSite(hostname, thirdPartyHostname)
3. Invent a new structure to pass to viz, this suffers from a similar issue to 1.
4. seems the most clean, however I am happy to merge this which I think currently is using 3. I think 2. would allow a little more flexibility for viz to draw much faster in the future and this code is in charge of what is drawn first.

js/viz.js
```diff
@@ -1,29 +1,111 @@
...
+      .selectAll('circle')
+      .data(nodesFirstParty)
+      .enter()
+      .append('circle')
+      .attr('r', 5)
+      .attr('fill', 'white');
+
+    // draw triangles for the thirdParty nodes
+    const nodeThirdParty = svg.append('g')
+      .attr('class', 'nodes')
```
biancadanforth: **jonathanKingston:**
multiline chained calls is fragile IMO does D3 advocate this by delaying rendering before it's complete?
I would prefer to see:
```javascript
const nodeThirdParty = svg.append('g');
nodeThirdParty.attr('class', 'nodes');
nodeThirdParty....
```
It also looks like the chaining is changing in return value? This makes it even more fragile.

js/viz.js
```diff
@@ -1,29 +1,111 @@
...
+      .data(linksData)
+      .enter().append('line')
+      .attr('stroke-width', 2);
+
+    // @todo draw text labels for each node
+
+    // update <circle> (x,y) positions to reflect node updates,
+    // update each <line> endpoint (x, y) position to reflect link updates
+    // on each tick of the simulation
+    function tickActions() {
```
biancadanforth: **jonathanKingston**:
Consider making this a method of viz rather than a nested function.

js/lightbeam.js
```diff
@@ -1,9 +1,33 @@
...
+    nodesData.push(node);
+    // make third party nodes and links
+    for (const thirdParty in websites[firstParty].thirdPartyRequests) {
+      const node = {};
+      const link = {};
+      link.source = firstParty;
+      link.target = thirdParty;
+      node.hostname = thirdParty;
+      // third parties likely don't have favicons
+      node.favicon = '';
```
biancadanforth: **jonathanKingston**:
This won't always be true, Google will likely.

js/lightbeam.js
```diff
@@ -1,9 +1,33 @@
...
+      node.party = 'third';
+      nodesData.push(node);
```
biancadanforth: **jonathanKingston**:
This code doesn't look like it accounts for multiple first parties having the same third party, won't it create multiple google analytics nodes for example?

index.html
```diff
@@ -88,7 +90,7 @@ <h1 id="filter">Recent Site</h1>
             <h2><span id="view">Graph</span> View</h2>
             <!-- @todo display list or graph view, update view based on filter and controls -->
             <!-- @todo remove the hard-coded width, height values -->
-            <canvas id="canvas" width="100%" height="100%"></canvas>
+            <svg width="700" height="700"></svg>
```
biancadanforth: **jonathanKngston**: Why not 100% again?

js/lightbeam.js
biancadanforth: I will try rewriting in the style of option 2. Stay tuned!

js/viz.js
biancadanforth: @jonathanKingston Unfortunately, per the d3 [documentation](https://github.com/d3/d3-force#simulation_on) for the `simulation.on(...)` method:
> When a specified event is dispatched, each listener will be invoked with the `this` context as the simulation.

Looking into it more, I found a better [explanation](https://stackoverflow.com/documentation/d3.js/2135/selections/27637/using-this-with-an-arrow-function#t=201703080935269580762):
> What can be useful to in several circumstances is a problem for a coder accustomed to use this in D3. For instance, using the same example in the fiddle above, this will not work:
```javascript
.on("mouseover", ()=>{
    d3.select(this);
});
```
In light of this, is there a way for me to write `tickActions` as a method on the `vis` object?

biancadanforth: Another [helpful article](https://medium.com/@yonester/on-d3-and-arrow-functions-b6559c1cebb8).

js/lightbeam.js
biancadanforth: No, it doesn't. I have fixed this with a `thirdParties` array that keeps track of third parties I that already have nodes. Only if a third party is not in that array will a new node be created.

js/viz.js
biancadanforth: I agree it makes things very confusing, as each method returns a new value that the subsequent method manipulates further. These are D3 methods though, and it looks like they designed their methods to be used in this chain fashion. Given that the value of `nodeThirdParty` is changed with each method, wouldn't I need to use `let` instead of `const` and have several re-assignment statements following the declaration?

```javascript
let nodeThirdParty = svg.append('g');
nodeThirdParty = nodeThirdParty.attr('class', 'nodes');
// ...
```
Why is this better in your view?

index.html
biancadanforth: The D3 library uses a property called `d3.forceCenter(width/2, height/2)` that centers the vis in the `<svg>` wrapper element. It does not accept percentage-based values and was defaulting to (0,0), so I reverted it to pixels again for the initial visualization to display properly.

I therefore need to compute the pixel height after the <svg> wrapper element for the vis has been rendered and pass that explicit (x,y) value in. Since I don't currently style this `<svg>` wrapper element in CSS, `element.clientWidth` and `element.clientHeight` return 0 (see [MDN](https://developer.mozilla.org/en-US/docs/Web/API/Element/clientWidth)), so I have used `element.getBoundingClientRect()`.

js/viz.js
jonathanKingston: > I need to use let instead of const and have several re-assignment statements following the declaration?

No not let, objects can be manipulated as I'm pretty sure I have seen you understand that. My understanding of your question is you are asking more about:

```
const nodeFirstParty = svg.append('g');
nodeFirstParty.attr('class', 'nodes')
const circles = nodeFirstParty.selectAll('circle')
circles.data(nodesFirstParty)
...
```

This is why I suggested this way, it's more explicit to someone not knowing d3 what the methods are operating on. (Also I'm guessing that is how the API works, it's not clear from the code at all as I don't know d3).

jonathanKingston: But if there is a perf argument to chaining though like it doesn't render until the last chain then I'm happy to ignore this.

biancadanforth:
For PR #67 , one thing that I feel I need to do is in-place modification of the nodesData array. This is done in two examples I've been referencing:
[example1](https://bl.ocks.org/tezzutezzu/cd04b3f1efee4186ff42aae66c87d1a7)
[example2](https://bl.ocks.org/mbostock/1095795)

js/lightbeam.js
```diff
@@ -1,9 +1,48 @@
...
-renderGraph();
+window.onload = async function() {
+  viz.init();
+  const websites = await store.getAll();
+  const data = transformData(websites);
+  // draw the elements onto the page
+  viz.draw(...data);
+};
+
+browser.storage.onChanged.addListener( (changes) => {
```
jonathanKingston: This shouldn't know how storage works. Please leave this out and we can work on this elsewhere.

js/lightbeam.js
biancadanforth: Based on [D3 documentation](https://bost.ocks.org/mike/join/) and my own experimentation, D3 has built-in methods that calculate the diff for a set of data and update the visualization without the need for us to tell it which nodes to add or remove (on a per-node or other basis). This PR demonstrates this. Happy to revisit; but this seems to be the way D3 expects us to pass in data (as a single array of objects).

js/lightbeam.js
biancadanforth: I created a follow-up issue #73 to fix this. Left you a question there as well.

js/viz.js
biancadanforth: No perf arguments that I could find. I removed the chaining.

js/lightbeam.js
biancadanforth: Okay.

jonathanKingston:
This is looking great, I just fixed `storeChild.getAll()` which should allow this to be plumbed into the real data when you get a chance.

Was this supposed to be using the mock data again? I don't really think we should be merging that.

biancadanforth:
I misunderstood you; I have removed the mock data!

jonathanKingston:
The diff still shows exmaple data, am I missing something?

biancadanforth:
Yes it was on platform 9 and 3/4. Sorry forgot to stage the changes before pushing the most recent changes...

js/lightbeam.js
```diff
@@ -1,9 +1,63 @@
...
+  let nodesPrev = data[0];
+  let linksPrev = data[1];
 
-  viz.draw(context, websites);
+  viz.init(nodesPrev, linksPrev);
+
+  // when data changes, update
+  // @todo find a good way to listen for changes
+  d3.interval(async function() {
+    websites = await storeChild.getAll();
```
jonathanKingston: Please remove this from this commit. Add the here code into a new bug so we can discuss.

This code shouldn't be interacting with d3 directly either doing so tightly couples this code to d3.

js/lightbeam.js
```diff
@@ -1,9 +1,63 @@
...
+
+  return [nodes, links];
+}
+
 async function renderGraph() {
-  const canvas = document.getElementById('canvas');
-  const context = canvas.getContext('2d');
-  const websites = await storeChild.getAll();
+  let websites = await storeChild.getAll();
+  let data = transformData(websites);
```
jonathanKingston: Consider returning keys from transformData.

```
let transformedData = transformData(websites);
viz.init(transformedData.nodes, transformedData.links);
```

js/lightbeam.js
```diff
@@ -1,9 +1,63 @@
+// transforms the object of nested objects 'websites' into a
+// usable format for d3
+function transformData(websites) {
+  const nodes = [];
+  const links = [];
+  const thirdParties = [];
```
jonathanKingston: Consider using maps, sets or objects here for thirdParties.
All provide a cleaner way to do:
```
if (thirdParties.indexOf(thirdParty) === -1) {
```

Could be:
```
if (thirdParties.has(thirdParty)) {
```

You could also instead just have:
```
function addWebsite(...) {
  // check I don't already exist yet
  // potentially merge favicons from third parties with first party nodes (Ie google.com)
  // some code
  nodes.push(node)
}
```

That way both first party and third party websites will be merged cleanly together. Currently I think you may end with a case of two nodes representing the same website.

We might also want to ditch the `"party"` key for the same reason. Google.com could be a first party and a third party to another website.

js/viz.js
```diff
@@ -1,29 +1,101 @@
...
+    this.simulation.force('charge', d3.forceManyBody());
+    const linkForce = d3.forceLink(links);
+    // link source/target values are uniquely identified by
+    // the hostname property
+    linkForce.id(function (d) {
+      return d.hostname;
+    });
+    linkForce.distance(50);
+    this.simulation.force('link', linkForce);
+    this.simulation.force('center', d3.forceCenter(width/2, height/2));
```
jonathanKingston: Is there any way to prevent this pinging off the graph. Mine zoomed out of sight pretty quickly... can be a follow up.

package.json
```diff
@@ -23,7 +23,9 @@
     "url": "https://github.com/electrolyfish/lightbeam-we/issues"
   },
   "homepage": "https://github.com/electrolyfish/lightbeam-we#readme",
-  "dependencies": {},
+  "dependencies": {
+    "d3": "4"
```
jonathanKingston: Use the full path here provided by: `npm install --save d3`

js/lightbeam.js
biancadanforth: Okay I did most of this (the `addWebsite` function and used `set` instead of arrays). It is true that Google could be a first and third party; I've added logic to ensure a third party node isn't created if a first party node already exists (but the link is still made).

The reason for the `party` key is to flag first parties as circles and third parties as triangles when d3 goes to draw the elements onto the page. How else might this distinction be made?

js/lightbeam.js
biancadanforth: I have taken this into consideration (and done this)!

Is this just because `data[0]` and `data[1]` is not as clear as using named object keys?

js/lightbeam.js
biancadanforth: I agree about that (also there is the built-in `setInterval` method which does the exact same thing). I have removed this portion from the code, so that currently the page only updates the visualization on page load.

@jonathanKingston , when you say make a bug, are you referring to the storage callback suggestion? Could you elaborate a bit on what that means? I created Issue #75.

js/viz.js
biancadanforth: There should be a way to do this. I made a follow-up issue. #76

package.json
biancadanforth: Okay; good idea.

js/lightbeam.js
```diff
@@ -1,9 +1,56 @@
...
+    firstParties.add(firstParty);
+    const node = {};
+    node.hostname = firstParty;
+    node.favicon = websites[firstParty].faviconUrl || '';
+    node.party = 'first';
+    nodes.push(node);
+    // make third party nodes and links
+    if (websites[firstParty].thirdPartyRequests) {
+      for (const thirdParty in websites[firstParty].thirdPartyRequests) {
+        addWebsite(firstParty, thirdParty);
```
jonathanKingston: Sorry I meant `addWebsite` would share code with first party in that it would for finding favicons etc not linking website nodes together.

So this code might look more like:

```
 for (const firstParty in websites) {
    firstParties.add(createWebsite(firstParty));
    // make third party nodes and links
    if (websites[firstParty].thirdPartyRequests) {
      for (const thirdParty in websites[firstParty].thirdPartyRequests) {
        firstParties.add(createWebsite(thirdParty));
        const link = {};
        link.source = firstParty.hostname;
        link.target = thirdParty.hostname;
        links.push(link);
```

```
function createWebsite(website) {
  const websiteOutput = {
    hostname: website.hostname,
    favicon: website.faviconUrl || "".
    party: "third"
  };
  if ("firstParties" in website) {
    websiteOutput.party = "first";
  }
  return websiteOutput;
}
```

js/lightbeam.js
```diff
@@ -1,9 +1,56 @@
...
+  for (const firstParty in websites) {
+    firstParties.add(firstParty);
+    const node = {};
+    node.hostname = firstParty;
+    node.favicon = websites[firstParty].faviconUrl || '';
+    node.party = 'first';
+    nodes.push(node);
+    // make third party nodes and links
+    if (websites[firstParty].thirdPartyRequests) {
+      for (const thirdParty in websites[firstParty].thirdPartyRequests) {
```
jonathanKingston: Is `thirdParty` here a hostname or is it an object? I thought it was an object.

css/style.css
```diff
@@ -1,11 +1,13 @@
...
-.circle {
-  display: inline-block;
-  width: 100px;
-  height: 100px;
-  border-radius: 50px;
-  background-color: red;
+.links line {
+  stroke: #999;
+  stroke-opacity: 0.6;
+}
```
jonathanKingston: Whitespace after the `}` that github doesn't show for some reason

js/lightbeam.js
biancadanforth: `thirdParty` is a hostname only; in fact, both `firstParty` and `thirdParty` that are part of the `for in` loops are strings. I did something along what you are suggesting!

js/lightbeam.js
biancadanforth: Okay I did something along these lines...

js/lightbeam.js
```diff
@@ -1,9 +1,52 @@
...
+  for (const firstParty in websites) {
+    const firstPartyNode = createWebsite(firstParty);
+    firstParties.add(firstPartyNode);
+    nodes.push(firstPartyNode);
+    // make third party nodes and links
+    if (websites[firstParty].thirdPartyRequests) {
+      for (const thirdParty in websites[firstParty].thirdPartyRequests) {
+        const thirdPartyNode = createWebsite(thirdParty);
+        thirdParties.add(thirdPartyNode);
+        if (!firstParties.has(thirdPartyNode)) {
```
jonathanKingston: This won't work, I thought we were just going to have a set of hostnames. Either way lets do this as a follow up which is fine.

jonathanKingston:
Merged code and fixed the whitespace from the css.
We should fix the `has` check code.

Issue #54, opened 2017-06-15
Storage api is populated with old values


princiya:
STR:

- Open Lightbeam
- Open console
- Clear the storage manually through the console: `browser.storage.local.remove('websites').then(()=>console.log('removed')).catch(err=>console.log(err))`
- Also set `store._websites=null` manually through the console
- Refresh the page - graph disappears
- Refresh the page again, graph appears with old values

This shouldn't be the case. We need to ensure there is only once instance of `store`.

jonathanKingston:
Unfortunately this is happening because there are more than one instance of store.js which is likely going to end in disaster.

The solution here is likely another object that messages to background version of store.js using message passing: https://developer.mozilla.org/en-US/Add-ons/WebExtensions/Content_scripts#Communicating_with_background_scripts

So one solution might look like:
```javascript
// store.js
const store = {
  async init() {
    const publicMethods = ["getAll"];
    browser.runtime.onMessage(responseHandler);
  },
  async responseHandler(m) => {
      if (publicMethods.includes(m.method)) {
        return this[m.method](...m.arguments);
      }
  },
  async getAll() {
    ...
  }
}
```

```javascript
// store-child.js
const storeChild = {
  getAll() {
    return await this.parentMessage("getAll", ...arguments);
  },
  parentMessage(method, ...arguments) {
    return browser.runtime.sendMessage({
      method,
      arguments
    })
  }
};
```

There might be something more elegant to make the code more DRY. Also naming might be something that can be improved, child/parent is used for processes within Firefox from content scripts to chrome level processes. However it's not really the same here.

The lightbeam page will use storeChild.js which will speak the the background script, the response from sendMessage is passed back as a promise, it should emulate the code pretty well as if it has the same instance.

princiya:
@jonathanKingston Thanks for the detailed explanation here :) Working on this now. I am glad we figured this disastrous thing during our initial stages!!!!!

jonathanKingston:
@princiya https://rosettacode.org/wiki/Respond_to_an_unknown_method_call#JavaScript can be used by the "store-child" to always call the parent and reduce overhead of having code in two places.

PR #70, opened 2017-06-22
Fix for the duplicate store

princiya:
Issue: #54 

As specified in the issue, we now have a `storeChild.js` which is the entry point for the `lightbeam.js` to talk to the background script `store.js`. Having this setup ensures there is always one `store` instance.

### DIFF
--- a/js/store.js
+++ b/js/store.js
@@ -14,6 +14,18 @@ const store = {
     }
   },
 
+  messageHandler(m) {
+    if (m.type !== 'storeCall') {
+      return;
+    }
+
+    const publicMethods = ['getAll'];
+
+    if (publicMethods.includes(m['method'])) {
+      return this[m['method']](...m.arguments);
+    }
+  },
+
   async _write(websites) {
     this._websites = clone(websites);
 
@@ -88,8 +100,6 @@ const store = {
   }
 };
 
-store.init();
-
 // @todo move this function to utils
 function clone(obj) {
   return Object.assign({}, obj);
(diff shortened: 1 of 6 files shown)


manifest.json
```diff
@@ -27,7 +27,6 @@
     "scripts": [
       "js/capture.js",
       "js/store.js",
-      "js/viz.js",
```
princiya: We don't need 'viz.js` to be a background script.

princiya:
I shall create a follow up PR to include the `[proxy](https://rosettacode.org/wiki/Respond_to_an_unknown_method_call#JavaScript)` object to avoid having same functions in both parent-child store objects.

Also, I plan to include all the other methods into `publicMethods`. This PR has only `getAll()`.

manifest.json
jonathanKingston: +1

js/store.js
```diff
@@ -93,5 +108,7 @@ function clone(obj) {
 function isObjectEmpty(obj) {
   return Object.keys(obj).length === 0;
 }
+// @todo end
```
jonathanKingston: Dunno what this is for

js/store.js
```diff
@@ -93,5 +108,7 @@ function clone(obj) {
...
 
 store.init();
+browser.runtime.onMessage.addListener(store.messageHandler.bind(store));
```
jonathanKingston: Please use arrows instead:
```
browser.runtime.onMessage.addListener(() => {
  store.messageHandler()
});
```

princiya:
@jonathanKingston used arrow functions.

removed unwanted code - `isEmptyObject()`. I cannot remember now why I had introduced this. I have done some code cleanup in this PR. Thanks for pointing it ou.

jonathanKingston:
@princiya I was actually asking about the `@todo` comment.

jonathanKingston:
Merging this for now. However `@todo` should be removed unless there is a reason for it.

PR #72, opened 2017-06-23
POC: d3 force layout with dynamic data

biancadanforth:
Got d3 to update without a re-render in a force-directed graph! Using fake data and setInterval currently.
The good news is that it does calculate the diff for us. We just have to pass in the updated value for nodes and links and preserve order by passing in a function that returns a unique ID for each node as the second argument for the `data()` method.
Also for a fraction of a second, some nodes appear at (0,0) (top left corner of the page) -- not sure if that is preventable but worth looking into as well.
Next step would be to draw with `<canvas>` instead and try using fake data that is even more similar to the form of data we would be passing in (maybe taking two or three snapshots of data and inserting it into this code)..

### DIFF
--- a/css/style.css
+++ b/css/style.css
@@ -1,11 +1,13 @@
 /* ----------- VISUALIZATION ----------- */
 
-.circle {
-  display: inline-block;
-  width: 100px;
-  height: 100px;
-  border-radius: 50px;
-  background-color: red;
+.links line {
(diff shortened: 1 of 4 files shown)


biancadanforth:
Closing this POC. It was just an experiment. Not intended to be merged.

PR #85, opened 2017-07-04
Storage callback for vis

jonathanKingston:


jonathanKingston:
Some of this is outdated code @biancadanforth we make the code asynchronous
in case it needs to be later through a different storage. Those vars have
been updated to website now from the latest update I think.
Updates come in per node so the code accounts for first party and third
party coming in out of order to the Lighbeam code.

js/lightbeam.js
```diff
@@ -2,51 +2,65 @@
...
+window.onload = initLightBeam;
 
+storeChild.onUpdate((data) => {
+  if (!(data.hostname in websites)) {
+    websites[data.hostname] = data;
+  }
+  if (!data.firstParty) {
+    // if we have the first party make the link
+    if (websites[data.firstPartyHostname]) {
+      const firstPartyWebsite = websites[data.firstPartyHostname];
```
biancadanforth: Per @jonathanKingston comment below:
> Updates come in per node so the code accounts for first party and third party coming in out of order to the Lighbeam code.

I don't really follow this answer... Perhaps I can walk through the code to better explain my confusion; I've added comments to explain what I think is going on..:

```javascript
// register callback function to execute when storage is updated; the callback is an anonymous arrow function
storeChild.onUpdate((data) => {
  // if the first or third party site is not already in websites, add it
  if (!(data.hostname in websites)) {
    websites[data.hostname] = data;
  }
  // if the site is a third party
  if (!data.firstParty) {
    // for each first party the third party links to
    data.firstPartyHostnames.forEach((firstPartyHostname) => {
      // if the first party exists in websites
      if (websites[firstPartyHostname]) {
        // capture its data in this new variable
        const firstPartyWebsite = websites[firstPartyHostname];
       // if the first party didn't previously have any linked third parties, initialize the thirdParty array
        if (!('thirdParties' in firstPartyWebsite)) {
          firstPartyWebsite.thirdParties = [];
          firstPartyWebsite.firstParty = true;
        }
        // if the first party does have third party connections, but it doesn't yet include this third party, then add this third party
        if (!(firstPartyWebsite.thirdParties.includes(data.hostname))) {
          firstPartyWebsite.thirdParties.push(data.hostname);
        }
      }
    });
  }
  // CONFUSION: nothing is done with this modified data object firstPartyWebsite?
  const transformedData = transformData(websites);
  viz.update(transformedData.nodes, transformedData.links);
});
```

jonathanKingston:
```
const myself = {shoe: 1, names: []};
let other = myself.shoe;
++other;
let names = myself.names;
names.push("beep");
console.log(myself);
```

Yeah I am lazy sometimes, basically we are depending on in place modifications of arrays/objects. Try running the above code, primitives are copied where as objects and arrays are not.

PR #95, opened 2017-07-10
Refactor storage shape to reduce lookups required for third parties.

jonathanKingston:
You can review it this time @princiya as I haven't had time to check this over much. Also this is your code and I want you to still be aware of it.

### DIFF
--- a/js/store.js
+++ b/js/store.js
@@ -45,17 +45,18 @@ const store = {
     return await browser.storage.local.set({ websites });
   },
 
-  outputWebsite(hostname, website, firstPartyHostname) {
+  outputWebsite(hostname, website) {
     const output = {
       hostname: hostname,
       favicon: website.faviconUrl || '',
-      firstPartyHostname: firstPartyHostname || false,
-      firstParty: false
+      firstPartyHostnames: website.firstPartyHostnames || false,
+      firstParty: false,
+      thirdParties: []
     };
-    if (firstPartyHostname === undefined) {
+    if (website.firstPartyHostnames === undefined) {
       output.firstParty = true;
-      if ('thirdPartyRequests' in website) {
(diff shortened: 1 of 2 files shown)


js/store.js
```diff
@@ -45,17 +45,17 @@ const store = {
...
 
-  outputWebsite(hostname, website, firstPartyHostname) {
+  outputWebsite(hostname, website) {
     const output = {
       hostname: hostname,
       favicon: website.faviconUrl || '',
-      firstPartyHostname: firstPartyHostname || false,
+      firstPartyHostnames: website.firstPartyHostnames || false,
       firstParty: false
     };
```
princiya: @jonathanKingston can we add `thirdParties: []` here and leave out the below `else` statement? In that way, the `output` object's signature would be complete.

js/store.js
```diff
@@ -45,17 +45,17 @@ const store = {
     return await browser.storage.local.set({ websites });
   },
 
-  outputWebsite(hostname, website, firstPartyHostname) {
+  outputWebsite(hostname, website) {
     const output = {
       hostname: hostname,
       favicon: website.faviconUrl || '',
-      firstPartyHostname: firstPartyHostname || false,
+      firstPartyHostnames: website.firstPartyHostnames || false,
```
princiya: Why is `firstPartyHostnames` an array now?

jonathanKingston: Because a third party can belong to multiple firsts.

js/store.js
jonathanKingston: The intention was the key wasn't present but either is fine.

js/store.js
jonathanKingston: The same thing you complained wasn't always accurate ;)

Issue #94, opened 2017-07-07
Two different graphs from onUpdate(capture) & page-refresh


princiya:
![two-graphs](https://user-images.githubusercontent.com/8022693/27976936-ddd16a04-6368-11e7-88c0-25cc1cd2824c.gif)

As seen from the above gif, currently we get two different graphs.

First one, from the capture phase or onUpdate event.
Second one, on page refresh.

We need to figure out what `nodes` are passed to the `viz` object during page reload and `onUpdate` stage.

princiya:
On debugging, I found that same nodes and links are passed during onUpdate and page-refresh. This issue is to do with the drawing of the graph and the x,y coordinates.

Issue #77, opened 2017-06-28
Allow scrolling of graph


jonathanKingston:
In the current light beam it is possible to scroll in the graph to zoom/in out.

Issue #92, opened 2017-07-05
Hover labels in graph


jonathanKingston:
Following from:
https://github.com/electrolyfish/lightbeam-we/pull/91

We would like to make the labels hover and also appear the same as they do in the current test pilot.

As we are likely to move to canvas this would have to position based on x and y on hover and look at the graph nodes position. (I'm not sure if d3 has this already somehow).

jonathanKingston:
@princiya did this... :)

PR #97, opened 2017-07-10
D3 updates

### LINKED ISSUE #76: Prevent subgraphs in visualization from going out of view
Currently, the forces on our force graph are causing some unattached nodes to fly out of view. Is there a way to confine the nodes to the bounds of the graph area?

princiya:
This PR has some updates to `viz.js`
- Tries to solve #94. Although this issue wasn't a bug, graphs are improved in this PR
- thirdParties are blue in color
- hover state for text labels

This PR aimed to solve #76, but as per discussions, we have decided to implement it in #77.

### DIFF
--- a/css/style.css
+++ b/css/style.css
@@ -12,6 +12,15 @@
 
 .textLabel {
   font-size: 14px;
+  visibility: hidden;
+}
+
+.textLabel.visible {
+  visibility: visible;
+}
+
+.textLabel.invisible {
+  visibility: hidden;
 }
 
 /* ----------- UI - General ----------- */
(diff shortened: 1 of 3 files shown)


princiya:
screen shot 2017-07-10 at 22 46 49

biancadanforth:
@princiya Tested this; still doesn't work. The `getX` and `getY` methods return the (x,y) coordinates (`width`, `height`) if the `d.x` exceeds `width` and `d.y` exceeds `height`, causing these nodes to collapse onto each other at (`width`, `height`).
Looking to see if there's a d3 solution -- or in general a way to know the bounding box dimensions  (and centroid) of each subgraph and adjust? Seems overly complicated...

princiya:
@jonathanKingston I have added hover state for labels in this PR. This might help @biancadanforth debug during the capture phase :)

This doesn't solve #92, because the styling isn't complete.

jonathanKingston:
Thanks @princiya!

PR #99, opened 2017-07-12
Add Disconnect Entity List (AKA whitelist). Fixes #98.

### LINKED ISSUE #98: Use the Disconnect Entity List to whitelist first-party-owned third parties

biancadanforth:
This PR includes the following changes:
* Add [Shavar-Prod-Lists](https://github.com/mozilla-services/shavar-prod-lists) repo as a submodule to use their maintained `disconnect-entitylist.json` whitelist. The submodule is located in the `shavar-prod-lists` subfolder in this repo's root directory.
* Updated `README.md` with new instructions for cloning the repo with the submodule and updating the submodule
* Import the `disconnect-entitylist.json` object, parsed, into `capture.js`
* Reformat the object to more easily check if a first and third party pair are whitelisted
* Update `capture.shouldStore` method to return `false` if the first and third party from a request are on the whitelist
  - @jonathanKingston: I couldn't use `object.hasOwnProperty` to check if the first party was in the whitelist, because the whitelist hostname is a substring of the first party request hostname (ex: "youtube.com" when I have "www.youtube.com" from the request). Any thoughts on how to improve the performance of this check would be greatly appreciated.
* Add `.eslintignore` to ignore submodule

Youtube.com subgraph before whitelist:
beforewhitelist
Youtube.com subgraph after whitelist:
afterwhitelist

### DIFF
--- a/js/capture.js
+++ b/js/capture.js
@@ -3,8 +3,16 @@
 * third-party requests to storage.
 */
 const capture = {
-  init() {
+  async init() {
     this.addListeners();
+    // get Disconnect Entity List from shavar-prod-lists submodule
+    let whiteList
+      = await fetch('/shavar-prod-lists/disconnect-entitylist.json');
+    whiteList = await whiteList.json();
+    const { firstPartyWhiteList, thirdPartyWhiteList }
+      = this.reformatList(whiteList);
+    this.firstPartyWhiteList = firstPartyWhiteList;
+    this.thirdPartyWhiteList = thirdPartyWhiteList;
   },
 
   addListeners() {
@@ -19,16 +27,125 @@ const capture = {
       });
   },
 
+/*
+  disconnect-entitylist.json is expected to match this format, where:
+    - 'properties' keys are first parties
+    - 'resources' keys are third parties
+
+  {
+    "Facebook" : {
+      "properties": [
+        "facebook.com",
+        "facebook.de",
+        ...
+        "messenger.com"
+      ],
+      "resources": [
+        "facebook.com",
+        "facebook.de",
+        ...
+        "akamaihd.net"
+      ]
+    }
+
+    "Google" : {
+      ...
+    }
+  }
+
+  this.firstPartyWhiteList is expected to match this format:
+  {
+    "google.com": 1,
+    "abc.xyz": 1
+    ....
+    "facebook.com": 2,
+    ...
+  }
+
+  this.thirdPartyWhiteList is expected to match this format:
+  {
+    1: [
+      "google.com",
+      "googleanalytics.com",
+      "weloveevilstuff.com"
+    ]
+  }
+*/
+
+  reformatList(whiteList) {
+    const firstPartyWhiteList = {};
+    const thirdPartyWhiteList = {};
+    let counter = 0;
+    for (const siteOwner in whiteList) {
+      const firstParties = whiteList[siteOwner].properties;
+      for (let i = 0; i < firstParties.length; i++) {
+        firstPartyWhiteList[firstParties[i]] = counter;
+      }
+      const thirdParties = whiteList[siteOwner].resources;
+      thirdPartyWhiteList[counter] = [];
+      for (let i = 0; i < thirdParties.length; i++) {
+        thirdPartyWhiteList[counter].push(thirdParties[i]);
+      }
+      counter++;
+    }
+
+    return {
+      firstPartyWhiteList,
+      thirdPartyWhiteList
+    };
+  },
+
   // Returns true if the request should be stored, otherwise false.
-  shouldStore(tab) {
+  shouldStore(tab, thirdPartyFromRequest) {
     const documentUrl = new URL(tab.url);
+    const firstPartyFromRequest = documentUrl.hostname;
     // ignore about:*, moz-extension:* & non-visible tabs (like dev tools)
+    // also ignore third parties owned by first parties
     if (documentUrl.protocol !== 'about:'
       && documentUrl.protocol !== 'moz-extension:'
-      && tab.id !== browser.tabs.TAB_ID_NONE) {
+      && tab.id !== browser.tabs.TAB_ID_NONE
+      && !(this.onWhitelist(firstPartyFromRequest, thirdPartyFromRequest))) {
       return true;
     }
     return false;
+  },
+
+  // check if third party is on the whitelist (owned by the first party)
+  // returns true if it is and false otherwise
+  onWhitelist(firstPartyFromRequest, thirdPartyFromRequest) {
+    if (thirdPartyFromRequest) {
+      const hostnameVariantsFirstParty
+        = this.getHostnameVariants(firstPartyFromRequest);
+      for (let i = 0; i < hostnameVariantsFirstParty.length; i++) {
+        if (this.firstPartyWhiteList
+          .hasOwnProperty(hostnameVariantsFirstParty[i])) {
+          // first party is in the whitelist
+          const index = this.firstPartyWhiteList[hostnameVariantsFirstParty[i]];
+          const hostnameVariantsThirdParty
+            = this.getHostnameVariants(thirdPartyFromRequest);
+          for (let j = 0; j < hostnameVariantsThirdParty.length; j++) {
+            if (this.thirdPartyWhiteList[index]
+              .includes(hostnameVariantsThirdParty[j])) {
+              return true;
+            }
+          }
+          return false;
+        }
+      }
+    }
+    return false;
+  },
+
+  getHostnameVariants(hostname) {
+    const hostnameVariants = [hostname];
+    const hostnameArr = hostname.split('.');
+    const numDots = hostnameArr.length - 1;
+    for (let i = 0; i < numDots - 1; i++) {
+      hostnameArr.shift();
+      hostname = hostnameArr.join('.');
+      hostnameVariants.push(hostname);
+    }
+    return hostnameVariants;
   },
 
   // capture third party requests
@@ -39,7 +156,7 @@ const capture = {
     const originUrl = new URL(response.originUrl);
 
     if (targetUrl.hostname !== documentUrl.hostname
-      && this.shouldStore(tab)) {
+      && this.shouldStore(tab, targetUrl.hostname)) {
       const data = {
         document: documentUrl.hostname,
         target: targetUrl.hostname,
--- a/.eslintignore
+++ b/.eslintignore
@@ -1,0 +1,1 @@
+/shavar-prod-lists/*
--- a/.gitmodules
+++ b/.gitmodules
@@ -1,0 +1,3 @@
+[submodule "shavar-prod-lists"]
+	path = shavar-prod-lists
+	url = https://github.com/mozilla-services/shavar-prod-lists.git
--- a/README.md
+++ b/README.md
@@ -5,11 +5,16 @@ The Mozilla Lightbeam extension is a key tool for Mozilla to educate the public 
 
 ## Quick Start
 
-* Clone the repository
-`git clone https://github.com/electrolyfish/lightbeam-we.git`
+### Clone the repository
 
-* There are a couple ways to try out this web extension:
-  1. Open Firefox and load `about:debugging` in the URL bar.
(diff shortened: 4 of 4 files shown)


js/capture.js
```diff
@@ -19,13 +27,61 @@ const capture = {
...
     const documentUrl = new URL(tab.url);
     // ignore about:*, moz-extension:* & non-visible tabs (like dev tools)
     if (documentUrl.protocol !== 'about:'
       && documentUrl.protocol !== 'moz-extension:'
       && tab.id !== browser.tabs.TAB_ID_NONE) {
+      // check if third party is on the whitelist (owned by the first party)
+      if (thirdPartyFromRequest) {
+        const firstPartyFromRequest = documentUrl.hostname;
+        for (const firstParty in this.firstPartyWhiteList) {
+          const firstPartyFromList = new RegExp(firstParty);
```
jonathanKingston: This won't do what you will want it to. Special characters in the firstParty var will break this matching correctly.

So:
youtube.co.uk would match: youtubesco.uk

We also might want to cater for sub domains but perhaps that might be better as a follow up? Or does it literally just match nothing? Perhaps the list could just test for firstParty and `www.${firstparty}`

js/capture.js
```diff
@@ -19,13 +27,61 @@ const capture = {
...
   // Returns true if the request should be stored, otherwise false.
-  shouldStore(tab) {
+  shouldStore(tab, thirdPartyFromRequest) {
     const documentUrl = new URL(tab.url);
     // ignore about:*, moz-extension:* & non-visible tabs (like dev tools)
     if (documentUrl.protocol !== 'about:'
       && documentUrl.protocol !== 'moz-extension:'
       && tab.id !== browser.tabs.TAB_ID_NONE) {
+      // check if third party is on the whitelist (owned by the first party)
+      if (thirdPartyFromRequest) {
```
jonathanKingston: So I think we should move this code outside of this function and instead have it tag the data.

Such that when the user loads:
- google.com

it loads a single node on the graph.
They then load:
- example.com

the graph should show google.com and example.com linking to google analytics.

js/capture.js
jonathanKingston: Potentially making this change also could to the object check too. Again check and let me know.

js/capture.js
```diff
@@ -19,13 +27,61 @@ const capture = {
       });
   },
 
+  // In the parsed whitelist JSON object (from disconnect-entitylist.json):
+  //   - 'properties' keys are first parties
+  //   - 'resources' keys are third parties
+  reformatList(whiteList) {
```
groovecoder: I don't always do it, but when I leave myself [a docblock showing an example of these lists' data formats](https://github.com/mozilla-services/shavar-prod-lists/blob/master/scripts/json_verify.py#L68-L102), I always thank myself later when I have to come back to understand how a function reformats or transforms this data.

In this case, it looks like the function turns this:
```
"Facebook" : {
  "properties": [
    "facebook.com",
    "facebook.de",
    ...
    "messenger.com"
  ],
  "resources": [
    "facebook.com",
    "facebook.de",
    ...
    "akamaihd.net"
  ]
}
```
into this:
```
{
  {
    "facebook.com": 0,
    "facebook.de": 1,
    ...
    "messenger.com": 9
  },
  {
    0: [ "facebook.com", "facebook.de", ... "akamaihd.com"],
    1: [ "facebook.com", "facebook.de", ... "akamaihd.com"],
    ...
    9: [ "facebook.com", "facebook.de", ... "akamaihd.com"]
  }
}
```
?

Whatever the structure of the data you make it (the logic could probably be optimized for performance and accuracy per @jonathanKingston's comments), I suggest adding a doc-block in front of this function to make it easy to remember later.

js/capture.js
biancadanforth: I agree with you about the regExp; I will need to escape the special characters (which should just be the decimal point, '.').

What do you mean by cater to subdomains? Are you suggesting we treat subdomains as different sites? Can you give me an example? There are a lot of nonstandard subdomains showing up in the third party hostnames from requests.

Also could you explain your last point in a little more detail?

js/capture.js
jonathanKingston: As this likely will be a follow up issue. We should move the algo below in this function to it's own method that this calls.

Then when we need to move this to store or somewhere else that is easier to do.

js/capture.js
jonathanKingston: Hmm well if you are just string matching why can't you just look up the thirdParty object key rather than comparing? Maybe the issue is also your data structure... let me check.

js/capture.js
jonathanKingston: Documentation is something I always miss... but yeah do this :D

js/capture.js
biancadanforth: I can move the code outside of the function. As for your second suggestion, per our slack discussion, this appears to warrant a follow-up issue. I will write one up shortly and reference it here.

js/capture.js
jonathanKingston: Nope format seems fine.

My understanding is with this code is using `RegExp` because you want to check if the hostname ends with something that is in the list.

Such that `google.com` would match `www.google.com` and `blah.google.com`. I'm not exactly sure if we want to do this for anything other than `www.` variants. Maybe @groovecoder can explain more how the subdomain matching works (I'm a little confused as some have `www.` but then goog doesn't).

jonathanKingston: The problem with this `RegExp` also is that it would also make `google.com` match for `google.com.arse.shoe.com` also which is an issue.

js/capture.js
biancadanforth: Great idea; made the change.

groovecoder: This is basically just how I code:
giphy

js/capture.js
biancadanforth: > Hmm well if you are just string matching why can't you just look up the thirdParty object key rather than comparing? Maybe the issue is also your data structure... let me check.

Do you mean why can't I use something like `whitelist.hasOwnProperty(firstParty)`? If so, it's because the first party string from the request is often not a direct match for what ought to be the corresponding string/property in the transformed whitelist object.

> Such that google.com would match www.google.com and blah.google.com. I'm not exactly sure if we want to do this for anything other than www. variants. 

But blah.google.com is google.com, so shouldn't it be subject to the same whitelist ignore check? Why would we only be interested in `www`?

> The problem with this RegExp also is that it would also make google.com match for google.com.arse.shoe.com also which is an issue.

I can modify the regExp to only look for cases where the substring `google.com` occurs at the end of the string in question. Would that solve it?

js/capture.js
jonathanKingston: Except Homer has way more hair.

js/capture.js
jonathanKingston: > Do you mean why can't I use something like whitelist.hasOwnProperty(firstParty)?

Yeah or `thirdParty.includes(thing)` for the other check.

> But blah.google.com is google.com, so shouldn't it be subject to the same whitelist ignore check? Why would we only be interested in www?

`www` is often treated as special for extensions like ours. I'm pretty confused why the list has `www.` and without if they are expecting different matching there. Like is it domain names or is it a hostname list?
I don't own the list so not sure, I would rather hope we could just duplicate `google.com` to `www.google.com` in the list and use an object key access and array includes rather than looping through all first parties if we can help it.

>  Would that solve it?

Technically no, it wouldn't account for left to right domains I don't think.
In order my preference:
- Key and array access
- string access modifying `www.google.com` to be `google.com` then comparing that inside the loop to array items
- constructing regexp at modification time
- improving what you have here

Either way if you use RegEx as you mentioned it will need to be escaped.

biancadanforth: > I would rather hope we could just duplicate google.com to www.google.com in the list and use an object key access and array includes rather than looping through all first parties if we can help it.

The hostnames from the third party requests are pretty gnarly. Examples include `r5---sn-4g5e6n76.googlevideo.com`, and the first parties' hostnames could be something like `developer.mozilla.com` with or without a subdomain. On the other end, first and third parties in the whitelist can be of the form `example.com` or `www.example.com`. I don't know all the possible variations on inputs from the requests and how to standardize these to use a direct string comparison using key and array access like `object.hasOwnProperty` or `array.includes`, particularly when the Disconnect list itself is inconsistent in its use of subdomains (`www` and even found a `www.hands.com.br`). That's why I used regExp.  If you have ideas, please share!

jonathanKingston: I really need confirmation from @groovecoder on what the list is all about. If it's hostnames or domain names.

I would suspect still that we can treat the both resources and properties as prefixed by www. and not. However we should get Luke to confirm. I will check the tracking protection code tomorrow in FF if not.

My real issue is domain handling isn't something I want to tackle in this project :P.

groovecoder: FWIW, [I used the string modification approach in `blok`](https://github.com/mozilla/blok/blob/4afc9d8bd7349a7b1c52a65e3b0d330648cebd47/src/js/lists.js#L108-L118). It added up to a few milliseconds processing time, but I forgave myself because I was already saving the user 5-10s of tracker processing time. 😉

In this context, I imagine that string processing won't be the performance bottleneck when compared to the visualization. You could start with string modification, and file an issue to refactor to something that performs better later. (Probably key & array access)

groovecoder:
Are the screenshots in the description here still accurate? It looks like many 3rd-party domains (e.g., parsely.com, cloudfront.net, __tising.com) are removed from the youtube.com graph? Those are certainly the 3rd-parties, while ytimg.com is actually 1st-party to Google. (It's the static thumbnail server for YT videos)

js/capture.js
biancadanforth: Per @jonathanKingston (paraphrasing)

Looks like the [safe browsing documentation](https://developers.google.com/safe-browsing/v4/urls-hashing) (the a.b.c example) suggests a basic solution to how to look at the lists:
* take developers.google.com and check every . into the name is in the list. Such that: developers.google.co.uk would check:
  - developers.google.co.uk
  - google.co.uk
  - co.uk

You pass through the list as many times as you have '.'s minus 1.
it's still (1) wrong for right to left and (2) overly expensive for companies that have extensions with .'s in them

What I mean by (1):
* hostnames are reversed in some languages
* assume this is in chinese: uk.co.thing.subdomain
* technically I think some hostnames can have mixed left to right and right to left which can screw up code too
* and there is also extensions that are right to left and hostnames that are left to right

What I mean by (2):
* as in .co.uk has an extra lookup
* some have 3 components too like .com.co.uk

That way you can use object/array lookups and only modify the URL coming in.

biancadanforth: @groovecoder Could you tell me what the incoming data types/structures are into the `hostInEntity` function you reference in your [example](https://github.com/mozilla/blok/blob/4afc9d8bd7349a7b1c52a65e3b0d330648cebd47/src/js/lists.js#L108-L118)?

biancadanforth: Per @groovecoder (paraphrase)
it’s the “resources” and “properties” objects part of the entity list. I pass them in along with `requestTopHost` `originTopHost` from [here](https://github.com/mozilla/blok/blob/4afc9d8bd7349a7b1c52a65e3b0d330648cebd47/src/js/requests.js#L36-L50).

biancadanforth:
@groovecoder ,

> Are the screenshots in the description here still accurate? It looks like many 3rd-party domains (e.g., parsely.com, cloudfront.net, __tising.com) are removed from the youtube.com graph? Those are certainly the 3rd-parties, while ytimg.com is actually 1st-party to Google. (It's the static thumbnail server for YT videos)

For those specific examples, I don't see those hostnames showing up at all from any of the requests picked up by `browser.onResponseStarted`, so they're never getting to the whitelist check in the first place. My guess is that youtube doesn't always make the exact same requests each time the home page loads? Though I loaded the page three times and never saw any of those sites you mention (ytimg excepted).

ytimg.com is not on the Disconnect Entity List, so that's why it's not being blocked.

groovecoder:
Okay yeah, I checked on my own and all the 3rd parties are definitely shown. :)

jonathanKingston:
@biancadanforth as mentioned I reverted this please can you check why mozilla isn't loading third parties please. https://developer.mozilla.org/en-US/docs/Tools/Web_Console is the link I was checking.

PR #101, opened 2017-07-14
D3 on canvas

### LINKED ISSUE #90: Migrate to using canvas d3 drawing
The current diagram is slowing down when lots of nodes are present. The graph should be migrated to using canvas to perform better.

princiya:
Fixes #90 

Canvas based solution for the visualisations.

PR started by following this blog: https://bocoup.com/blog/d3js-and-canvas
As pointed out in this blog, the idea is to create custom elements which aren't loaded into the DOM, but are in memory (virtual dom) and then repainting the canvas. Doing so, we get to use D3's core algorithms (force layout in our case).

We don't use dummy elements anymore. D3's force layout algorithm is used and nodes, links are painted on the canvas.

### DIFF
--- a/css/style.css
+++ b/css/style.css
@@ -290,6 +290,11 @@ input[type="checkbox"] {
   color: var(--secondary-text-color);
 }
 
+#visualization {
+  width: 500px;
+  height: 400px;
+}
+
 .info-panel-controls {
   display: flex;
   flex-direction: column;
--- a/js/viz.js
+++ b/js/viz.js
@@ -5,35 +5,38 @@ const viz = {
 
   init(nodes, links) {
     const { width, height } = this.getDimensions('visualization');
-    const { context, custom } = this.createCanvas(width, height);
 
     this.width = width;
     this.height = height;
-    this.context = context;
-    this.custom = custom;
-    this.circles = custom.selectAll('custom.circle');
-    this.labels = custom.selectAll('custom.label');
-    this.lines = custom.selectAll('custom.line');
+    this.context = this.createCanvas(width, height);
 
     this.draw(nodes, links);
   },
 
-  simulate(nodes, links) {
-    this.simulation = this.simulateForce(nodes, links);
+  draw(nodes, links) {
+    this.nodes = nodes;
+    this.links = links;
+
+    this.simulate();
+    this.drawOnCanvas();
+  },
+
+  simulate() {
+    this.simulation = this.simulateForce();
     this.updatePositions();
     this.simulation.tick();
   },
 
-  simulateForce(nodes, links) {
-    const linkForce = d3.forceLink(links);
+  simulateForce() {
+    const linkForce = d3.forceLink(this.links);
     let simulation;
 
     if (!this.simulation) {
-      simulation = d3.forceSimulation(nodes);
+      simulation = d3.forceSimulation(this.nodes);
     } else {
       simulation = this.simulation;
       this.simulation.stop();
-      simulation.nodes(nodes);
+      simulation.nodes(this.nodes);
     }
 
     linkForce.id((d) => d.hostname);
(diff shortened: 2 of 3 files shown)


css/style.css
```diff
@@ -290,6 +290,11 @@ input[type="checkbox"] {
   color: var(--secondary-text-color);
 }
 
+#visualization {
+  width: 500px;
```
princiya: @todo: remove these hard-coded values

princiya:
At this point, I don't fully understand D3's force layout algorithm. Once this piece is fully understood, canvas based optimisations are possible too.

Hover for text labels isn't there in this PR. Interactivity using canvas is a little tricky, so I will do that in the next PR. @jonathanKingston

princiya:
Canvas based visualisations and the custom elements which are in virtual dom. 

screen shot 2017-07-14 at 22 10 18

js/viz.js
```diff
@@ -59,33 +72,51 @@ const viz = {
   },
 
   updatePositions() {
-    this.allCircles
-      .attr('cx', (d) => d.x)
-      .attr('cy', (d) => d.y);
-    this.allLabels
+    this.circles
```
princiya: we are not using SVGs, so this piece isn't needed too. However at this point I don't know how to update the coordinates. So I leave this as it is.

princiya:
@biancadanforth @jonathanKingston I have never done performance testing using devtools. Tried something today, for a given sample of websites and here are the results. I have no idea how to interpret these results :-/

Old Lightbeam:

screen shot 2017-07-15 at 17 11 55

New Lightbeam - SVG:

screen shot 2017-07-15 at 17 11 47

New Lightbeam - Canvas:

screen shot 2017-07-15 at 17 11 32

One thing is certain, the load time for canvas is the fastest ;)

js/viz.js
jonathanKingston: I'm confused the circles still seem to move, what is the issue with this code?

jonathanKingston:
I'm shocked that adding those custom nodes is any faster. Let me read the article it would be better not to use invalid elements and also not to use DOM at all as this will bloat over time.

js/viz.js
```diff
@@ -98,38 +129,39 @@ const viz = {
...
+  },
 
-    this.simulation = this.simulateForce(nodes, links);
-    this.updatePositions();
-    this.simulation.tick();
+  draw(nodes, links) {
+    this.simulate(nodes, links);
+    this.createVirtualDom(nodes, links);
+    this.drawOnCanvas();
+    this.simulate(nodes, links);
```
jonathanKingston: Why is simulate called twice in this method?

js/viz.js
```diff
@@ -98,38 +129,39 @@ const viz = {
...
-    let newText = this.allLabels.enter();
-    newText = newText.append('text');
-    newText.attr('class', (d, i) => `textLabel text${i}`);
-    newText.text((d) => d.hostname);
-    newText.attr('fill', 'white');
-
-    this.allLabels = newText.merge(this.allLabels);
+  drawLinks() {
+    this.context.beginPath();
+    this.lines.each((d) => {
```
jonathanKingston: Links or lines? Can we have them named the same?

js/viz.js
```diff
@@ -98,38 +129,39 @@ const viz = {
...
+    });
+    this.context.closePath();
+    this.context.strokeStyle = '#ccc';
+    this.context.stroke();
   },
 
-  drawLinks(links) {
-    this.allLines = this.allLines
-      .data(links, (d) => `${d.source.hostname}-${d.target.hostname}`);
+  virtualDom(type, elements) {
```
jonathanKingston: This isn't virtual, if we are keeping with this method then rename this to be about being fake dom perhaps.

jonathanKingston:
Why do updates look so radically different now?

Graph update:
selection_716
Refreshing the page:
selection_717

I noticed that sometimes on many updates the graph gets completely off screen too.

jonathanKingston:
It looks like the reason they end up off the graph is the two subgraphs repel each other which hapens with more force than normal when many third parties are at once.

STR:
- Load mozilla property
- Load google, click all their top links fast
- Canvas has no contents

jonathanKingston:
We should try and ensure that two disconnected subgraphs don't ever overlay over each other. It looks way better in the update view (when they are not off screen).

PR #104, opened 2017-07-17
Revert "Add Disconnect Entity List (AKA whitelist). Fixes #98."

### LINKED ISSUE #98: Use the Disconnect Entity List to whitelist first-party-owned third parties

jonathanKingston:
Reverts electrolyfish/lightbeam-we#99

### DIFF
--- a/.eslintignore
+++ b/.eslintignore
@@ -1,1 +1,0 @@
-/shavar-prod-lists/*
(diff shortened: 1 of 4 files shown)


jonathanKingston:
@biancadanforth I just loaded the patch and it loads no third parties for https://developer.mozilla.org could you investigate before we merge please? thanks

PR #106, opened 2017-07-18
Add Disconnect entity list. Fixes #98.

### LINKED ISSUE #98: Use the Disconnect Entity List to whitelist first-party-owned third parties

biancadanforth:
This PR includes the following changes:
Add [Shavar-Prod-Lists](https://github.com/mozilla-services/shavar-prod-lists) repo as a submodule to use their maintained `disconnect-entitylist.json` whitelist. The submodule is located in the `shavar-prod-lists` subfolder in this repo's root directory.
Update `README.md` with new instructions for cloning the repo with the submodule and updating the submodule
Import the `disconnect-entitylist.json` object, parsed, into `capture.js`
Reformat the object to more easily check if a first and third party pair are whitelisted
Update `capture.shouldStore` method to return `false` if the first and third party from a request are on the whitelist
Add `.eslintignore` to ignore submodule
**NEW**: Fix `store.setWebsite` method so that it doesn't silently fail when `this._websites` is `null`.

### DIFF
--- a/README.md
+++ b/README.md
@@ -5,11 +5,16 @@ The Mozilla Lightbeam extension is a key tool for Mozilla to educate the public 
 
 ## Quick Start
 
-* Clone the repository
-`git clone https://github.com/electrolyfish/lightbeam-we.git`
+### Clone the repository
 
-* There are a couple ways to try out this web extension:
-  1. Open Firefox and load `about:debugging` in the URL bar.
+**Note** This repository uses a [submodule](https://github.com/mozilla-services/shavar-prod-lists) to whitelist some third party requests. To ensure the submodule is cloned along with this repository, use a modified `clone` command:
+`git clone --recursive https://github.com/electrolyfish/lightbeam-we.git`
+
+### Run the web extension
+
+There are a couple ways to try out this web extension:
+
+1. Open Firefox and load `about:debugging` in the URL bar.
     - Click the [Load Temporary Add-on](https://developer.mozilla.org/en-US/Add-ons/WebExtensions/Temporary_Installation_in_Firefox) button and select the `manifest.json` file within the directory of this repository.
     - You should now see the Lightbeam icon on the top right bar of the browser.
     - Click the Lightbeam icon to launch the web extension.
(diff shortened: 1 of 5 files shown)


biancadanforth:
This is to unrevert [this PR](https://github.com/biancadanforth/lightbeam-we/commit/8a1b4ba00cc629f708f8c519cb1f55cf0436e030).

See that PR for an explanation of the additional fix here, "Fix `store.setWebsite` method so that it doesn't silently fail when `this._websites` is `null`".

js/capture.js
```diff
@@ -3,8 +3,16 @@
 * third-party requests to storage.
 */
 const capture = {
-  init() {
+  async init() {
     this.addListeners();
+    // get Disconnect Entity List from shavar-prod-lists submodule
+    let whiteList
+      = await fetch('/shavar-prod-lists/disconnect-entitylist.json');
```
jonathanKingston: Add error handler here.

js/capture.js
```diff
@@ -19,18 +27,127 @@ const capture = {
...
+      && !(this.onWhitelist(firstPartyFromRequest, thirdPartyFromRequest))) {
       return true;
     }
     return false;
   },
 
+  // check if third party is on the whitelist (owned by the first party)
+  // returns true if it is and false otherwise
+  onWhitelist(firstPartyFromRequest, thirdPartyFromRequest) {
+    if (thirdPartyFromRequest) {
```
jonathanKingston: Check firstPartyWhiteList is not undefined here, that way if fetch fails we should be safe. The code can just return false if we don't have a whitelist. I would rather not block capture ever either.

js/capture.js
```diff
@@ -3,8 +3,16 @@
 * third-party requests to storage.
 */
 const capture = {
-  init() {
+  async init() {
     this.addListeners();
```
jonathanKingston: The other solution to the whitelist not working is move this after the whitelist has loaded. This means we shouldn't block capturing by the entity list.

If the fetch fails perhaps make the lists empty arrays?

README.md
```diff
@@ -5,33 +5,43 @@ The Mozilla Lightbeam extension is a key tool for Mozilla to educate the public
...
 ## Development Guide
 
-* Run `npm install`.
-* Run `npm run test` to check that everything is OK.
-  * If you have installed `eslint` globally, you will have to install globally the following `eslint` plugins too:
+### Download dependencies
+Run `npm install`.
+
+### Update the submodule
+To manually update the submodule at any time during development, run `git submodule update`.
```
jonathanKingston: Maybe mention `git submodule init` for new users perhaps?

js/capture.js
biancadanforth: Ok; using `.then` chaining currently; what options do I have to do this? You mentioned using an absolute URL as a web extension resource -- how would I determine that? Submitting current form which uses `.then`.

js/capture.js
biancadanforth: Okay done

js/capture.js
biancadanforth: I updated the code so that an error is printed to the console about the json file not being fetched properly, but the code still works (just sans whitelist).

README.md
biancadanforth: New users would use the modified `git clone --recursive https://github.com/electrolyfish/lightbeam-we.git` which would take care of init (referring to this [GitHub guide on submodules](https://github.com/blog/2104-working-with-submodules). They would only need an `init` command if they did not do the recursive clone command from the README.

Issue #102, opened 2017-07-17
Interactivity on canvas


princiya:
We now have `canvas` for visualisations. #101 

Interactivity is a little tricky on canvas. This issue is to setup the base for creating interactions on canvas. To start with, let's make the labels interactive!

We had an issue for interactive labels #92, this was an SVG based solution.

PR #108, opened 2017-07-19
Graph labels on canvas

princiya:
Issue: #102 

Canvas label screenshot

### DIFF
--- a/js/viz.js
+++ b/js/viz.js
@@ -1,12 +1,27 @@
 // eslint-disable-next-line no-unused-vars
 const viz = {
+  radius: 15,
+
   init(nodes, links) {
     const { width, height } = this.getDimensions('visualization');
+    const { canvas, context } = this.createCanvas({
+      width,
+      height
+    });
 
     this.width = width;
     this.height = height;
-    this.context = this.createCanvas(width, height);
+    this.canvas = canvas;
+    this.context = context;
+    /* this.hiddenContext = this.createCanvas({
+      width,
+      height,
+      display: 'none'
+    }); */
+    // colors to node mapping for canvas based mouse interactivity
+    this.colToNode = {};
 
+    this.addListeners();
     this.draw(nodes, links);
   },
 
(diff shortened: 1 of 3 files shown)


princiya:
@jonathanKingston need your help :)

onUpdate, tooltips work fine (please ignore the circle radius)

Labels

onPageRefresh, tooltips don't work. not able to figure out, what's wrong.

Labels

js/viz.js
```diff
@@ -114,5 +132,48 @@ const viz = {
...
+    this.addMouseMove();
+    // this.addMouseLeave();
+  },
+
+  addMouseMove() {
+    this.canvas.on('mousemove', () => {
+      const points = d3.mouse(this.canvas.node());
+      // const X = d3.event.layerX || d3.event.offsetX;
+      const node = this.getNodeAtCoordinates(points[0], points[1]);
+      // const node = this.simulation.find(points[0], points[1], this.radius);
```
princiya: `simulation.find()` returns the node closest to the position ⟨x,y⟩ with the given search radius.
`getNodeAtCoordinates()` is what I have written to find out if <x,y> is inside a given circle.
Both of the above implementations fail on page refresh, and don't return any node most of the time.

js/viz.js
```diff
@@ -1,13 +1,22 @@
 // eslint-disable-next-line no-unused-vars
 const viz = {
+  radius: 5,
```
jonathanKingston: Maybe call this websiteRadius or something more descriptive?

js/viz.js
```diff
@@ -114,5 +132,48 @@ const viz = {
...
+  },
+
+  addListeners() {
+    this.addMouseMove();
+    // this.addMouseLeave();
+  },
+
+  addMouseMove() {
+    this.canvas.on('mousemove', () => {
+      const points = d3.mouse(this.canvas.node());
```
jonathanKingston: You can use destructuring here to add clarity:
```
const [mouseX, mouseY] = d3.mouse(this.canvas.node());
```

js/viz.js
jonathanKingston: My hunch is that this would be right if another tick happens. We perhaps need to store currentX and Y of nodes until tick happens. Try calling tick first before calculating and see if it works. I have a feeling we have conflated the meaning of x and y here.

jonathanKingston: Or if we are not doing that perhaps we just need to call tick more?

princiya:
The labels now work on page refresh too. @jonathanKingston 

Here is the logic:

Because we would expect lot of nodes in our graph, it's best to do a manual tick rather than registering an event handler for tick.

From the [d3-force](https://github.com/d3/d3-force) docs:
tick

Before the graph is drawn, we compute simulations and set the `x,y` coordinates for the nodes & links. Next, `simulation.stop()` is invoked to compute a static force layout in conjunction with `simulation.tick()`.

On page refresh, because we missed this `simulation.stop()`, `simulation.tick()` had updated node values but the graph was drawn at the old values.

Issue #65, opened 2017-06-19
Make vars dynamic


jonathanKingston:
From #61

Make `<var` and `<time>` elements dynamic from storage data.

biancadanforth:
@jonathanKingston Here is some brainstorming about how to go about this:

**How to determine each value:**
1. Data gathered since
 - If there's nothing in the store, (i.e.: if it's the first run of the app or if the user clicks 'Reset Data'), update this value to be blank or '--'
 - else, the value is equal to the date of the oldest `website` object in the `websites` table in IndexedDB.
 - only needs to be updated on init/'Reset Data'
2. You have visited (# visited sites)
- If there's nothing in the store, this value is 0.
- else, the value is equal to the number of `website` objects in store with a `firstParty` of `true`.
- Should be updated with each new first party addition to the `websites` store
3. You have connected with (# third party sites)
2. You have visited (# visited sites)
- If there's nothing in the store, this value is 0.
- else, the value is equal to the number of `website` objects in store with a `firstParty` of `false`. AKA the total number of websites in the store minus those that are first parties.
- - Should be updated with each new third party addition to the `websites` store

**Where to store this:**
Make a separate store to the `websites` one we currently have, since the structure of the data for these dynamic variables is different.

jonathanKingston:
We could instead populate the page async based on the current data in store
when the find result returns.

Changing store to ensure we have first capture time would potentially be
easier and useful for removing old data later.

princiya:
> How to determine each value:
Looks good to me.

> Where to store this:
I don't think you need a separate store. We already have this information in our store.
We can have a separate page-script like `viz.js`, something like `vars.js` (name can be changed).

*`var.js`*
```
const vars = {
 date: '',
 firstParty: '',
 thirdParty: '',

 init() {
  // populate the above vars by querying the store 
 }
}
```

The way `viz.js` gets input about websites from store, `vars.js` needs to be populated too.
The only thing to take care of is how to refresh the `visited` count when new first/third parties are added.
We might need a listener which constantly checks for new values and updates them.

jonathanKingston:
I think these vars that princiya mentions could be stored and updated in
lightbeam.js

Also to be clear "websites" should be able to store this data.

biancadanforth:
@jonathanKingston Why wouldn't we want to keep the data in a separate store from `websites`? The UI variables don't have any of the same keys as the `website` objects in `websites`.

Here is the structure of the `websites` database currently:

[...]

PR #109, opened 2017-07-20
Override whitelist for special case. Fixes #100.

### LINKED ISSUE #100: Override whitelist when a third party links to multiple first parties
A paraphrased example from @jonathanKingston:

Say you have two sites that both link to google-analytics.com:
1. google.com (owns google-analytics.com)
2. example.com (does not own google-analytics.com)

Initially, say you have only loaded google.com, the graph should only display the first party node for google.com, not the google-analytics.com third party node, since google-analytics.com is owned by google.com (i.e. google-analytics.com is whitelisted per #98 ).

When you visit example.com next, currently the graph would only show the link between example.com and google-analytics.com. There would not be a link between google.com and google-analytics.com due to the whitelist check. We should override this to display this link in this instance.

To put it more generally, if a previously whitelisted third party later links to one or more first parties for which it is not whitelisted, the previously ignored link between the whitelisted first and third party should be added.

In addition to adding more transparency around the prevalence of a particular third party site, this has the added benefit of connecting two subgraphs, which can help keep the visualization more contained to the graph area in the UI (Issue #76).

@jonathanKingston Could you please elaborate on how this could be done, and what it might take implementation-wise?
@groovecoder Do you know why some hostnames in the [Disconnect Entity List](https://github.com/mozilla-services/shavar-prod-lists/blob/master/disconnect-entitylist.json) have the `www` subdomain and others don't?

biancadanforth:
Add Doc Block for `store.setThirdParty` method and `lightbeam.js` `transformData` function
Move whitelist logic from `capture.js` to `store.js`
Add override logic inside `store.setThirdParty` with `isIgnored` boolean

Also, I used these two websites for my test cases: tylersavin.com and gravatar.com. Both link to `s.gravatar.com`, which is owned by gravatar.com

@jonathanKingston Still needs some refactoring in `store.setThirdParty` to breakout the whitelist code there. Working on that next. Any advice welcome.

### DIFF
--- a/js/store.js
+++ b/js/store.js
@@ -1,5 +1,6 @@
 const store = {
   _websites: null,
+  WHITELIST_URL: '/shavar-prod-lists/disconnect-entitylist.json',
 
   async init() {
     if (!this._websites) {
@@ -10,18 +11,20 @@ const store = {
       }
     }
     browser.runtime.onMessage.addListener((m) => store.messageHandler(m));
+    await this.getWhiteList();
+  },
 
-    // get Disconnect Entity List from shavar-prod-lists submodule
+  // get Disconnect Entity List from shavar-prod-lists submodule
+  async getWhiteList() {
     let whiteList;
-    const whiteListURL = '/shavar-prod-lists/disconnect-entitylist.json';
     try {
-      whiteList = await fetch(whiteListURL);
+      whiteList = await fetch(this.WHITELIST_URL);
       whiteList = await whiteList.json();
     } catch (error) {
       whiteList = {};
       const explanation = 'See README.md for how to import submodule file';
       // eslint-disable-next-line no-console
-      console.error(`${error.message} ${explanation} ${whiteListURL}`);
+      console.error(`${error.message} ${explanation} ${this.WHITELIST_URL}`);
     }
     const { firstPartyWhiteList, thirdPartyWhiteList }
       = this.reformatList(whiteList);
@@ -152,11 +155,31 @@ const store = {
     const output = {};
     for (const hostname in this._websites) {
       const website = this._websites[hostname];
-      output[hostname] = this.outputWebsite(hostname, website);
+      if (website['isVisible']) {
+        output[hostname] = this.outputWebsite(hostname, website);
+      }
     }
     return Promise.resolve(output);
   },
 
+  /*
+  if not {}, getWebsite returns an object:
+  - in this format for first parties:
+    {
+      thirdPartyHostnames: [
+        "www.thirdpartydomain.com",
+        ...
+      ]
+    }
+  and in this format for third parties:
+    {
+      firstPartyHostnames: [
+        "www.firstpartydomain.com",
+        ...
+      ],
+      isVisible: false,
+    }
+  */
   getWebsite(hostname) {
     if (!hostname) {
       throw new Error('getWebsite requires a valid hostname argument');
@@ -182,12 +205,9 @@ const store = {
   },
 
   setWebsite(hostname, data) {
-    let newSite = false;
-
     const websites = clone(this._websites);
 
-    if (!websites[hostname]) {
-      newSite = true;
+    if (this.isNewWebsite(hostname)) {
       websites[hostname] = {};
     }
 
@@ -196,10 +216,14 @@ const store = {
     }
 
     this._write(websites);
+  },
 
-    if (newSite && !data['isIgnored']) {
-      this.updateChild(this.outputWebsite(hostname, websites[hostname]));
+  isNewWebsite(hostname) {
+
+    if (!this._websites || !this._websites[hostname]) {
+      return true;
     }
+    return false;
   },
 
   getHostnameVariants(hostname) {
@@ -244,94 +268,85 @@ const store = {
     if (!hostname) {
       throw new Error('setFirstParty requires a valid hostname argument');
     }
+    const isNewWebsite = this.isNewWebsite(hostname);
+
     this.setWebsite(hostname, data);
+
+    if (isNewWebsite) {
+      this.updateChild(this.outputWebsite(hostname, this._websites[hostname]));
+    }
   },
 
+  // @todo use data from capture to update UI/filter graph
+  // eslint-disable-next-line no-unused-vars
   setThirdParty(origin, target, data) {
-    let newThirdParty = false;
+    let isNewWebsite = false;
     if (!origin) {
       throw new Error('setThirdParty requires a valid origin argument');
     }
 
-    const firstParty = this.getWebsite(origin);
-    const thirdParty = this.getWebsite(target);
-
-/*
-  if not {},
-
-  firstParty is expected to match this format:
-  {
-    thirdPartyHostnames: [
-      "www.thirdpartydomain.com",
-      ...
-    ]
-  }
-  thirdParty is expected to match this format:
-  {
-    firstPartyHostnames: [
-      "www.firstpartydomain.com",
-      ...
-    ],
-    isIgnored: true, (optional)
-    whiteListedFirstParty: "www.google.com" (optional)
-  }
-  */
-
-    let whiteListOverride = false;
-    let onWhiteList = false;
+    let firstParty = this.getWebsite(origin);
+    let thirdParty = this.getWebsite(target);
 
     if (!('thirdPartyHostnames' in firstParty)) {
       firstParty['thirdPartyHostnames'] = [];
     }
     if (!firstParty['thirdPartyHostnames'].includes(target)) {
       // third party is not currently linked to first party
-      if (!thirdParty['isIgnored']) {
-        // third party needs a whitelist check
-        onWhiteList = this.onWhiteList(origin, target);
-        if (onWhiteList) {
-          // third party is whitelisted for this first party
-          // ignore the link for now
-          thirdParty['isIgnored'] = true;
-          thirdParty['whiteListedFirstParty'] = origin;
-        } else {
-          // third party is not whitelisted for this first party
-          firstParty['thirdPartyHostnames'].push(target);
-          newThirdParty = true;
-        }
-      } else if (!(this.onWhiteList(origin, target))) {
-        // third party was previously whitelisted, and its
-        // previously ignored link with the owning first party
-        // should be added now that it links to a first party
-        // for which it is not whitelisted
-        thirdParty['isIgnored'] = false;
-        whiteListOverride = true;
-        const whiteListedFirstParty = thirdParty['whiteListedFirstParty'];
-        this._websites[whiteListedFirstParty]['thirdPartyHostnames']
-          .push(target);
-        firstParty['thirdPartyHostnames'].push(target);
-        // take care of updating thirdParty[firstPartyHostnames] further down
-        newThirdParty = true;
-      }
+      ({ firstParty, thirdParty, isNewWebsite }
+        = this.whiteListCheck(origin, target, firstParty, thirdParty));
     }
     if (!('firstPartyHostnames' in thirdParty)) {
       thirdParty['firstPartyHostnames'] = [];
     }
     if (!thirdParty['firstPartyHostnames'].includes(origin)) {
-      if (whiteListOverride) {
-        thirdParty['firstPartyHostnames']
-          .push(thirdParty['whiteListedFirstParty']);
-      }
-      if (newThirdParty) {
-        thirdParty['firstPartyHostnames'].push(origin);
-      }
+      thirdParty['firstPartyHostnames'].push(origin);
     }
 
-    this.setWebsite(origin, firstParty);
+    this.setFirstParty(origin, firstParty);
     this.setWebsite(target, thirdParty);
 
-    if (newThirdParty) {
-      this.updateChild(this.outputWebsite(target, data, origin));
+    if (isNewWebsite && thirdParty['isVisible']) {
+      this.updateChild(this.outputWebsite(target, this._websites[target]));
     }
+  },
+
+  // Checks if third party is on whitelist for first party,
+  // and if the whitelisted third party should be unwhitelisted.
+  whiteListCheck(origin, target, firstParty, thirdParty) {
+    let isNewWebsite = false;
+    if (!('isVisible' in thirdParty)) {
+      if (this.onWhiteList(origin, target)) {
+        // third party is whitelisted for this first party
+        // ignore this third party for now.
+        thirdParty['isVisible'] = false;
+      } else {
+        thirdParty['isVisible'] = true;
+        firstParty['thirdPartyHostnames'].push(target);
+        isNewWebsite = true;
+      }
+    } else if (!(this.onWhiteList(origin, target))) {
+      if (!thirdParty['isVisible']) {
+        // third party was previously whitelisted, and its
+        // previously ignored link with the owning first party
+        // should be added now
+        thirdParty['isVisible'] = true;
+        const whiteListedFirstParty = thirdParty['firstPartyHostnames']
+          .filter((firstPartyHostname) => {
+            return !(this._websites[firstPartyHostname]['thirdPartyHostnames']
+              .includes(target));
+          })[0];
+        this._websites[whiteListedFirstParty]['thirdPartyHostnames']
+          .push(target);
+      }
+      firstParty['thirdPartyHostnames'].push(target);
+      isNewWebsite = true;
+    }
+    return {
+      firstParty,
+      thirdParty,
+      isNewWebsite
+    };
   },
 
   async reset() {
--- a/js/store.js
+++ b/js/store.js
@@ -205,15 +205,19 @@ const store = {
     return null;
   },
 
-  setWebsite(hostname, data) {
+  setWebsite(hostname, partyData, requestData) {
     const websites = clone(this._websites);
 
     if (this.isNewWebsite(hostname)) {
       websites[hostname] = {};
     }
 
-    for (const key in data) {
-      websites[hostname][key] = data[key];
+    for (const key in partyData) {
+      websites[hostname][key] = partyData[key];
+    }
+
+    for (const key in requestData) {
+      websites[hostname][key] = requestData[key];
     }
 
     this._write(websites);
@@ -285,7 +289,7 @@ const store = {
     }
 
     let isNewThirdParty = false;
-    let isNewFirstParty = false;
+    let shouldUpdate = false;
 
     const firstParty = this.getWebsite(origin);
     const thirdParty = this.getWebsite(target);
@@ -296,49 +300,52 @@ const store = {
     }
     if (!thirdParty['firstPartyHostnames'].includes(origin)) {
       thirdParty['firstPartyHostnames'].push(origin);
-      isNewFirstParty = true;
     }
 
     // add link in first party if it doesn't exist yet
     // and the third party is visible (i.e. not whitelisted)
-    if (!('thirdPartyHostnames' in  firstParty)
-      || !firstParty['thirdPartyHostnames'].includes(target)) {
-      if (!('isVisible' in thirdParty) || !thirdParty['isVisible']) {
-        if (!(this.onWhiteList(origin, target))) {
-          // show third party
-          thirdParty['isVisible'] = true;
-          isNewThirdParty = true;
-        } else {
+    if (!this.isFirstPartyConnectedToThirdParty(firstParty, target)) {
+      if (!this.isVisibleThirdParty(thirdParty)) {
+        if (this.onWhiteList(origin, target)) {
           // hide third party
           thirdParty['isVisible'] = false;
+        } else {
+          // show third party and all its connections
+          thirdParty['isVisible'] = true;
+          isNewThirdParty = true;
+          thirdParty['firstPartyHostnames'].forEach((firstPartyHostname) => {
+            this.addFirstPartyLink(firstPartyHostname, target);
+          });
+          shouldUpdate = true;
         }
       }
-      if (thirdParty['isVisible'] && isNewFirstParty) {
+      if (this.isVisibleThirdParty(thirdParty) && !isNewThirdParty) {
         // an existing visible third party links to a new first party
-        thirdParty['firstPartyHostnames'].forEach((firstPartyHostname) => {
-          this.addFirstPartyLink(firstPartyHostname, target);
-        });
+        this.addFirstPartyLink(origin, target);
+        shouldUpdate = true;
       }
     }
 
-    // merge data with thirdParty website object
-    for (const key in data) {
-      thirdParty[key] = data[key];
-    }
-
-    this.setWebsite(target, thirdParty);
-
-    // should update if:
-    //   - there is a new third party, OR
-    //   - there is a new link between an old third party and a new first party
-    //   AND
-    //   - the third party is visible (i.e. not whitelisted)
-    const shouldUpdate = (isNewThirdParty || isNewFirstParty)
-      && thirdParty['isVisible'];
+    this.setWebsite(target, thirdParty, data);
 
     if (shouldUpdate) {
       this.updateChild(this.outputWebsite(target, thirdParty));
     }
+  },
+
+  isFirstPartyConnectedToThirdParty(firstParty, thirdPartyHostname) {
+    if (!('thirdPartyHostnames' in  firstParty)
+      || !firstParty['thirdPartyHostnames'].includes(thirdPartyHostname)) {
+      return false;
+    }
+    return true;
+  },
+
+  isVisibleThirdParty(thirdParty) {
+    if (!('isVisible' in thirdParty) || !thirdParty['isVisible']) {
+      return false;
+    }
+    return true;
   },
 
   addFirstPartyLink(firstPartyHostname, thirdPartyHostname) {
@@ -346,8 +353,11 @@ const store = {
     if (!('thirdPartyHostnames' in firstParty)) {
       firstParty['thirdPartyHostnames'] = [];
     }
-    firstParty['thirdPartyHostnames'].push(thirdPartyHostname);
-    this.setFirstParty(firstPartyHostname, firstParty);
+    if (!this
+      .isFirstPartyConnectedToThirdParty(firstParty, thirdPartyHostname)) {
+      firstParty['thirdPartyHostnames'].push(thirdPartyHostname);
+      this.setFirstParty(firstPartyHostname, firstParty);
+    }
   },
 
   async reset() {
--- a/js/store.js
+++ b/js/store.js
@@ -1,4 +1,3 @@
-// eslint-disable-next-line no-unused-vars
 const store = {
   _websites: null,
 
@@ -11,6 +10,91 @@ const store = {
       }
     }
     browser.runtime.onMessage.addListener((m) => store.messageHandler(m));
+
+    // get Disconnect Entity List from shavar-prod-lists submodule
+    let whiteList;
+    const whiteListURL = '/shavar-prod-lists/disconnect-entitylist.json';
+    try {
+      whiteList = await fetch(whiteListURL);
+      whiteList = await whiteList.json();
+    } catch (error) {
+      whiteList = {};
+      const explanation = 'See README.md for how to import submodule file';
+      // eslint-disable-next-line no-console
+      console.error(`${error.message} ${explanation} ${whiteListURL}`);
+    }
+    const { firstPartyWhiteList, thirdPartyWhiteList }
+      = this.reformatList(whiteList);
+    this.firstPartyWhiteList = firstPartyWhiteList;
+    this.thirdPartyWhiteList = thirdPartyWhiteList;
+  },
+
+  /*
+  disconnect-entitylist.json is expected to match this format, where:
+    - 'properties' keys are first parties
+    - 'resources' keys are third parties
+
+  {
+    "Facebook" : {
+      "properties": [
+        "facebook.com",
+        "facebook.de",
+        ...
+        "messenger.com"
+      ],
+      "resources": [
+        "facebook.com",
+        "facebook.de",
+        ...
+        "akamaihd.net"
+      ]
+    }
+
+    "Google" : {
+      ...
+    }
+  }
+
+  this.firstPartyWhiteList is expected to match this format:
+  {
+    "google.com": 1,
+    "abc.xyz": 1
+    ....
+    "facebook.com": 2,
+    ...
+  }
+
+  this.thirdPartyWhiteList is expected to match this format:
+  {
+    1: [
+      "google.com",
+      "googleanalytics.com",
+      "weloveevilstuff.com"
+    ]
+  }
+*/
+
+  reformatList(whiteList) {
+    const firstPartyWhiteList = {};
+    const thirdPartyWhiteList = {};
+    let counter = 0;
+    for (const siteOwner in whiteList) {
+      const firstParties = whiteList[siteOwner].properties;
+      for (let i = 0; i < firstParties.length; i++) {
+        firstPartyWhiteList[firstParties[i]] = counter;
+      }
+      const thirdParties = whiteList[siteOwner].resources;
+      thirdPartyWhiteList[counter] = [];
+      for (let i = 0; i < thirdParties.length; i++) {
+        thirdPartyWhiteList[counter].push(thirdParties[i]);
+      }
+      counter++;
+    }
+
+    return {
+      firstPartyWhiteList,
+      thirdPartyWhiteList
+    };
   },
 
   // send message to storeChild when data in store is changed
@@ -113,9 +197,47 @@ const store = {
 
     this._write(websites);
 
-    if (newSite) {
+    if (newSite && !data['isIgnored']) {
       this.updateChild(this.outputWebsite(hostname, websites[hostname]));
     }
+  },
+
+  getHostnameVariants(hostname) {
+    const hostnameVariants = [hostname];
+    const hostnameArr = hostname.split('.');
+    const numDots = hostnameArr.length - 1;
+    for (let i = 0; i < numDots - 1; i++) {
+      hostnameArr.shift();
+      hostname = hostnameArr.join('.');
+      hostnameVariants.push(hostname);
+    }
+    return hostnameVariants;
+  },
+
+  // check if third party is on the whitelist (owned by the first party)
+  // returns true if it is and false otherwise
+  onWhiteList(firstPartyFromRequest, thirdPartyFromRequest) {
+    if (thirdPartyFromRequest && this.firstPartyWhiteList) {
+      const hostnameVariantsFirstParty
+        = this.getHostnameVariants(firstPartyFromRequest);
+      for (let i = 0; i < hostnameVariantsFirstParty.length; i++) {
+        if (this.firstPartyWhiteList
+          .hasOwnProperty(hostnameVariantsFirstParty[i])) {
+          // first party is in the whitelist
+          const index = this.firstPartyWhiteList[hostnameVariantsFirstParty[i]];
+          const hostnameVariantsThirdParty
+            = this.getHostnameVariants(thirdPartyFromRequest);
+          for (let j = 0; j < hostnameVariantsThirdParty.length; j++) {
+            if (this.thirdPartyWhiteList[index]
+              .includes(hostnameVariantsThirdParty[j])) {
+              return true;
+            }
+          }
+          return false;
+        }
+      }
+    }
+    return false;
   },
 
   setFirstParty(hostname, data) {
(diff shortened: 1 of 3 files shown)


js/store.js
```diff
@@ -11,6 +10,91 @@ const store = {
       }
     }
     browser.runtime.onMessage.addListener((m) => store.messageHandler(m));
+
+    // get Disconnect Entity List from shavar-prod-lists submodule
+    let whiteList;
```
princiya: @biancadanforth best to move this whitelist related part into a separate function. its nice to keep in `init` only variable assignments and further function calls.

js/store.js
```diff
@@ -134,19 +256,74 @@ const store = {
...
+    ],
+    isIgnored: true, (optional)
+    whiteListedFirstParty: "www.google.com" (optional)
+  }
+  */
+
+    let whiteListOverride = false;
+    let onWhiteList = false;
+
     if (!('thirdPartyHostnames' in firstParty)) {
```
princiya: @biancadanforth is possible, better to move this whitelist related code too into a separate function. I don't know the functionality yet, but since `setFirstParty` has too many lines, it is better to break them into functions.

js/store.js
jonathanKingston: Yea separate function would be a little cleaner. Feel free to put hard coded contants at the top of the file too with long var names. Like `const WHITELIST_BLAH_SOMETHING_URL = ....`

js/store.js
jonathanKingston: This is Github trolling you... it collapsed the method name. I was confused too.

js/store.js
```diff
@@ -134,19 +256,74 @@ const store = {
...
+        const whiteListedFirstParty = thirdParty['whiteListedFirstParty'];
+        this._websites[whiteListedFirstParty]['thirdPartyHostnames']
+          .push(target);
+        firstParty['thirdPartyHostnames'].push(target);
+        // take care of updating thirdParty[firstPartyHostnames] further down
+        newThirdParty = true;
+      }
     }
-    if (!('firstPartyHostnames' in firstParty)) {
+    if (!('firstPartyHostnames' in thirdParty)) {
```
jonathanKingston: Indeed this was a bug.

js/store.js
```diff
@@ -134,19 +256,74 @@ const store = {
...
       thirdParty['firstPartyHostnames'] = [];
     }
     if (!thirdParty['firstPartyHostnames'].includes(origin)) {
-      thirdParty['firstPartyHostnames'].push(origin);
-      newThirdParty = true;
+      if (whiteListOverride) {
+        thirdParty['firstPartyHostnames']
+          .push(thirdParty['whiteListedFirstParty']);
```
jonathanKingston: What is the difference between: `thirdParty['whiteListedFirstParty']` and `origin`? Looks to me like it is always origin?

jonathanKingston: My hunch is that we always want to store `origin` in this list anyways.

js/store.js
jonathanKingston: That comment above was to @princiya.

@princiya is right though in that once the algo is right we could split it into other functions prehaps. I'm still reading through the logic though.

js/store.js
```diff
@@ -134,19 +256,74 @@ const store = {
...
+    let onWhiteList = false;
+
     if (!('thirdPartyHostnames' in firstParty)) {
       firstParty['thirdPartyHostnames'] = [];
     }
     if (!firstParty['thirdPartyHostnames'].includes(target)) {
-      firstParty['thirdPartyHostnames'].push(target);
-      newThirdParty = true;
+      // third party is not currently linked to first party
+      if (!thirdParty['isIgnored']) {
```
jonathanKingston: Change the name to something truthy rather than false. not not is hard to parse for most people. So far it looks like you currenty only check ! anyway for this property.

So the statement might be:
```
if (thirdParty['isVisible']) {
```

jonathanKingston: `isIgnored` actually seems wrong on second read.

js/store.js
```diff
@@ -134,19 +256,74 @@ const store = {
...
+        // third party needs a whitelist check
+        onWhiteList = this.onWhiteList(origin, target);
+        if (onWhiteList) {
+          // third party is whitelisted for this first party
+          // ignore the link for now
+          thirdParty['isIgnored'] = true;
+          thirdParty['whiteListedFirstParty'] = origin;
```
jonathanKingston: Technically a third party might appear on multiple first parties. I don't think it makes sense to store this.

js/store.js
```diff
@@ -134,19 +256,74 @@ const store = {
...
+        } else {
+          // third party is not whitelisted for this first party
+          firstParty['thirdPartyHostnames'].push(target);
+          newThirdParty = true;
+        }
+      } else if (!(this.onWhiteList(origin, target))) {
```
jonathanKingston: Why not else? is the return value for `onWhiteList` not a bool?

js/store.js
```diff
@@ -134,19 +256,74 @@ const store = {
...
+      "www.firstpartydomain.com",
+      ...
+    ],
+    isIgnored: true, (optional)
+    whiteListedFirstParty: "www.google.com" (optional)
+  }
+  */
+
+    let whiteListOverride = false;
+    let onWhiteList = false;
```
jonathanKingston: Don't initialise this, or even have this at all?

js/store.js
jonathanKingston: Overall I think this code would be much simpler if:

- you store state of isIgnored/isHidden/isVisible or whatever var name you choose
- you only return websites that are visible in `getAll`
- you keep storing firstPartyHostnames in third party as this will never be seen until visible
- when this code decides it becomes visible, it should make sure all websites in `firstPartyHostnames`  have a reference to the thirdParty in their `thirdPartyHostnames` arrays.
- when a website becomes visible the update should fire, the update should not fire for this website until then

js/store.js
biancadanforth: Okay great suggestion thank you. Is there a way to consolidate the code below into a single line?

```javascript
    const { firstPartyWhiteList, thirdPartyWhiteList }
      = await this.getWhiteList();
    this.firstPartyWhiteList = firstPartyWhiteList;
    this.thirdPartyWhiteList = thirdPartyWhiteList;
```
I tried this,

```javascript
{ this.firstPartyWhiteList this.thirdPartyWhiteList } = await this.getWhiteList();
```
But I get a `SyntaxError: expected expression, got '='` message in the console...

biancadanforth: Ah I thought of one approach that worked:
```javascript
await this.getWhiteList();
```
And have `this.getWhiteList` assign to `this.firstPartyWhiteList` and `this.thirdPartyWhiteList` instead of returning values to the `this.init` method.

biancadanforth: @jonathanKingston , for putting constants at the top, do you mean something like this:

```javascript
const store = {
  _websites: null,
  WHITELIST_URL: '/shavar-prod-lists/disconnect-entitylist.json',

  async init() {
    // ...
```
Or outside of the `store` object entirely? Wouldn't we want to keep it inside `store` so as not to "pollute the global namespace"?

js/store.js
biancadanforth: Okay I changed the boolean key to `isVisible`, good point on double negatives.

Also agreed, I had it reversed, which is wrong. Fixed this.

js/store.js
biancadanforth: I have these booleans for this bit of code further down `store.setThirdParty` to update the `thirdParty` website object based on the whitelist/override check done when the `firstParty website object was updated, so I don't have to perform the whitelist check again.


```javascript
if (!('firstPartyHostnames' in thirdParty)) {
      thirdParty['firstPartyHostnames'] = [];
    }
    if (!thirdParty['firstPartyHostnames'].includes(origin)) {
      if (whiteListOverride) {
        thirdParty['firstPartyHostnames']
          .push(thirdParty['whiteListedFirstParty']);
      }
      if (newThirdParty) {
        thirdParty['firstPartyHostnames'].push(origin);
      }
    }
```

It's possible with some refactoring of this method that I wouldn't need these booleans.

js/store.js
biancadanforth: Do you mean a third party might be whitelisted for multiple first parties?

I have to store this whitelisted first party because otherwise I lose which first party caused the third party to be ignored when a new third party request comes in.

In the case that the new third party request requires me to make visible the previously ignored third party, I need to know the original first party hostname that whitelisted the third party, so that I can update the third party's object (`firstPartyHostnames` key) and whitelisted first party's object (`thirdPartyHostnames` key) in `this_websites`.

I could make it an array instead perhaps?

js/store.js
biancadanforth: `thirdParty['whiteListedFirstParty']` is only the same as `origin` in the initial third party request for which the third party is whitelisted by the request's first party.

In this if statement, `origin` is a new first party for which the third party is not whitelisted, so `origin` is not equal to `thirdParty['whiteListedFirstParty']`.

js/store.js
biancadanforth: > you store state of isIgnored/isHidden/isVisible or whatever var name you choose

This is done already.

> you only return websites that are visible in getAll

This is now done:
```javascript
  getAll() {
    const output = {};
    for (const hostname in this._websites) {
      const website = this._websites[hostname];
      // only previously ignored websites will have an 'isVisible' key
      if (website['isVisible'] || !('isVisible' in website)) {
        output[hostname] = this.outputWebsite(hostname, website);
      }
    }
    return Promise.resolve(output);
  },
```



> you keep storing firstPartyHostnames in third party as this will never be seen until visible

Good point; this is done. Reverted this update back to its original logic:
```javascript
    if (!thirdParty['firstPartyHostnames'].includes(origin)) {
      thirdParty['firstPartyHostnames'].push(origin);
    }
```
> when this code decides it becomes visible, it should make sure all websites in firstPartyHostnames have a reference to the thirdParty in their thirdPartyHostnames arrays.

This is already done:
```javascript
// inside store.setThirdParty
} else if (!(this.onWhiteList(origin, target))) {
        // third party was previously whitelisted, and its
        // previously ignored link with the owning first party
        // should be added now that it links to a first party
        // for which it is not whitelisted
        // ... other stuff
        const whiteListedFirstParty = thirdParty['whiteListedFirstParty'];
        this._websites[whiteListedFirstParty]['thirdPartyHostnames']
          .push(target);
        // other stuff...
      }
```

This then gets written to storage when `setWebsite` is called.

> when a website becomes visible the update should fire, the update should not fire for this website until then

This is already done.

`newThirdParty` is only set to `true` if:
1. The third party is not whitelisted for the current request
2. The previously whitelisted third party has linked to a new first party for which it is not whitelisted

```javascript
// in store.setThirdParty
    if (newThirdParty) {
      this.updateChild(this.outputWebsite(target, data, origin));
    }
```

There is also a check in `store.setWebsite`:

```javascript
if (newSite && data['isVisible']) {
      this.updateChild(this.outputWebsite(hostname, websites[hostname]));
    }
```

js/store.js
```diff
@@ -157,6 +306,37 @@ const store = {
...
+      // should be added now that it links to a first party
+      // for which it is not whitelisted
+      thirdParty['isVisible'] = true;
+      const whiteListedFirstParty = thirdParty['whiteListedFirstParty'];
+      this._websites[whiteListedFirstParty]['thirdPartyHostnames']
+        .push(target);
+      firstParty['thirdPartyHostnames'].push(target);
+      newThirdParty = true;
+    }
+    return [firstParty, thirdParty, newThirdParty];
```
jonathanKingston: Can you make this an object instead, argument return order is a little fragile:
```
return {firstParty, thirdParty, newThirdParty};
```

js/store.js
```diff
@@ -157,6 +306,37 @@ const store = {
...
+    let newThirdParty = false;
+    if (!('isVisible' in thirdParty) || thirdParty['isVisible']) {
+      // third party needs a whitelist check
+      if (this.onWhiteList(origin, target) && !('isVisible' in thirdParty)) {
+        // third party is whitelisted for this first party
+        // ignore the link for now
+        thirdParty['isVisible'] = false;
+        thirdParty['whiteListedFirstParty'] = origin;
+      } else {
+        // third party is not whitelisted for this first party
```
jonathanKingston: In this else statement shouldn't `isVisible` be set to true?

js/store.js
```diff
@@ -157,6 +306,37 @@ const store = {
...
+  // and if the whitelisted third party should be unwhitelisted.
+  whiteListCheck(origin, target, firstParty, thirdParty) {
+    let newThirdParty = false;
+    if (!('isVisible' in thirdParty) || thirdParty['isVisible']) {
+      // third party needs a whitelist check
+      if (this.onWhiteList(origin, target) && !('isVisible' in thirdParty)) {
+        // third party is whitelisted for this first party
+        // ignore the link for now
+        thirdParty['isVisible'] = false;
+        thirdParty['whiteListedFirstParty'] = origin;
```
jonathanKingston: Please just make `firstPartyHostnames` consistent for `isVisible` or not and when it becomes visible loop through `firstPartyHostnames` to clear up to add it to `thirdPartyHostnames` as the website might exist to multiple whitelisted first parties which currently would be a bug.

js/store.js
```diff
@@ -68,7 +155,10 @@ const store = {
     const output = {};
     for (const hostname in this._websites) {
       const website = this._websites[hostname];
-      output[hostname] = this.outputWebsite(hostname, website);
+      // only previously ignored websites will have an 'isVisible' key
```
jonathanKingston: I'm pretty sure the above statement could be made not true. As soon as a third party is known not to have a first party on the whitelist it can have the key.
If we simplify this to only ever be two states we should make the code more reliable.

js/store.js
```diff
@@ -113,11 +203,49 @@ const store = {
 
     this._write(websites);
 
-    if (newSite) {
+    if (newSite && data['isVisible']) {
```
jonathanKingston: If the previous statement is true about this key not always being present this would be a bug here.

js/store.js
biancadanforth: Well, it's not possible in this version of the code. For some reason (which escapes me now), I felt I needed to have the key only exist on `thirdParty` if it was initially ignored. I will make the change to have the `isVisible` key always present for a `thirdParty` object.

js/store.js
biancadanforth: Yes you're right; making that key always present should allow this to work (see previous comment above).

js/store.js
biancadanforth: I initially tried to do that; but then I get a `Parsing error: Unexpected token =` error where the `this.whiteListCheck` is called:

```javascript
      { firstParty, thirdParty, newThirdParty }
        = this.whiteListCheck(origin, target, firstParty, thirdParty);
```
Any ideas why? Also why are argument return orders more consistent using an object instead of an array? I guess this isn't really an object since it doesn't have key-value pairs...

js/store.js
biancadanforth: Yes using the updated logic that `isVisible` is present for all `thirdParty` website objects. Previously I had only added the key if a third party was to be ignored.

js/store.js
biancadanforth: Done; thank you. Learned a bit about destructuring assignments in the process.

js/store.js
```diff
@@ -131,32 +258,95 @@ const store = {
...
+      this.updateChild(this.outputWebsite(target, data));
     }
   },
 
+  // Checks if third party is on whitelist for first party,
+  // and if the whitelisted third party should be unwhitelisted.
+  whiteListCheck(origin, target, firstParty, thirdParty) {
+    let newThirdParty = false;
+    // first parties are always visible
+    if (!('isVisible' in firstParty)) {
```
jonathanKingston: I don't think we need this key always present in firstParty do we? Especially as this won't be called for first parties when they are new this would be unreliable.

js/store.js
```diff
@@ -131,32 +258,95 @@ const store = {
       throw new Error('setThirdParty requires a valid origin argument');
     }
 
-    const firstParty = this.getWebsite(origin);
-    const thirdParty = this.getWebsite(target);
+    let firstParty = this.getWebsite(origin);
+    let thirdParty = this.getWebsite(target);
+
+/*
```
jonathanKingston: Can this big comment be placed outside of the function?

biancadanforth: Sure; done.

js/store.js
biancadanforth: No we don't. Fixed this by putting the call to `storeChild.update` into `store.setFirstParty` and `store.setThirdParty`.

js/store.js
```diff
@@ -112,51 +216,139 @@ const store = {
...
-      newThirdParty = true;
     }
 
-    this.setWebsite(origin, firstParty);
+    this.setFirstParty(origin, firstParty);
     this.setWebsite(target, thirdParty);
 
-    if (newThirdParty) {
-      this.updateChild(this.outputWebsite(target, data, origin));
+    if (isNewWebsite && thirdParty['isVisible']) {
```
jonathanKingston: I think we can simplify this further by making this one function. I don't think there is an obvious way to make this a shorter function without moving some of the initialising our into the set or get functions perhaps but for now I think this is fine:

Please test this (perhaps in a branch from this branch):

```js
setThirdParty(origin, target, data) {
  if (!origin) { 
    throw new Error('setThirdParty requires a valid origin argument');
  } 

  let isNewThirdParty = false;
  let firstParty = this.getWebsite(origin);
  let thirdParty = this.getWebsite(target);

  if (!('firstPartyHostnames' in thirdParty)) { 
    thirdParty['firstPartyHostnames'] = [];
  } 
  if (!thirdParty['firstPartyHostnames'].includes(origin)) { 
    thirdParty['firstPartyHostnames'].push(origin);
  } 

  // third party is not currently linked to first party
  if (!firstParty['thirdPartyHostnames'].includes(target)) { 
    /* When the third party isn't visible or we are not sure,
       we check the whitelist */
    if (!('isVisible' in thirdParty)
        || !thirdParty.isVisible) { 
      /* When we find a first party not on the whitelist we choose to show the third party,
         which requires sending out updates to anything listening */
      if (!this.onWhiteList(origin, target)) { 
        thirdParty['isVisible'] = true;
        isNewThirdParty = true;
        thirdParty['firstPartyHostnames'].forEach((firstPartyHostname) => { 
          this.addFirstPartyRelationship(firstPartyHostname, target);
        });
      } 
    } 
  } 

  this.setWebsite(target, thirdParty);

  if (isNewThirdParty && thirdParty['isVisible']) { 
    this.updateChild(this.outputWebsite(target, this._websites[target]));
  } 
},

addFirstPartyRelationship(firstPartyHostname, thirdPartyHostname) {
  const firstPartyInstance = this.getWebsite(firstPartyHostname);
  if (!('thirdPartyHostnames' in firstPartyInstance)) { 
    firstPartyInstance['thirdPartyHostnames'] = [];
  } 
  if (!firstParty['thirdPartyHostnames'].includes(target)) { 
    firstPartyInstance['thirdPartyHostnames'].push(target);
    this.setFirstParty(firstPartyHostname, firstPartyInstance);
  }
}
```

js/store.js
```diff
@@ -112,51 +216,139 @@ const store = {
...
+      if (this.onWhiteList(origin, target)) {
+        // third party is whitelisted for this first party
+        // ignore this third party for now.
+        thirdParty['isVisible'] = false;
+      } else {
+        thirdParty['isVisible'] = true;
+        firstParty['thirdPartyHostnames'].push(target);
+        isNewWebsite = true;
+      }
+    } else if (!(this.onWhiteList(origin, target))) {
```
jonathanKingston: This is in expensive order, we should check if it's not visible first, See my code snippet.

js/store.js
```diff
@@ -112,51 +216,139 @@ const store = {
...
+      if (!thirdParty['isVisible']) {
+        // third party was previously whitelisted, and its
+        // previously ignored link with the owning first party
+        // should be added now
+        thirdParty['isVisible'] = true;
+        const whiteListedFirstParty = thirdParty['firstPartyHostnames']
```
jonathanKingston: We aren't setting stuff here for the first parties, which is an issue if the site has more than one connection already that was whitelisted: `['abc.xyz', 'google.com'...]` for example. See my code snippet.

js/store.js
```diff
@@ -112,51 +216,139 @@ const store = {
...
+        this._websites[whiteListedFirstParty]['thirdPartyHostnames']
+          .push(target);
+      }
+      firstParty['thirdPartyHostnames'].push(target);
+      isNewWebsite = true;
+    }
+    return {
+      firstParty,
+      thirdParty,
+      isNewWebsite
```
jonathanKingston: When state like this is tracked across two functions it's a code smell. Sure we do some of this in the visualisation code however it's not a great common practice. The setThirdParty code has some of the logic for what these vars might mean which is why I said the bracket assign syntax isn't common: 
`({ firstParty, thirdParty, isNewWebsite }...`

js/store.js
```diff
@@ -112,51 +216,139 @@ const store = {
...
   setFirstParty(hostname, data) {
     if (!hostname) {
       throw new Error('setFirstParty requires a valid hostname argument');
     }
+    const isNewWebsite = this.isNewWebsite(hostname);
+
     this.setWebsite(hostname, data);
+
+    if (isNewWebsite) {
+      this.updateChild(this.outputWebsite(hostname, this._websites[hostname]));
```
jonathanKingston: Why isn't `this._websites[hostname]` just data instead?

js/store.js
```diff
@@ -112,51 +216,139 @@ const store = {
...
     }
 
-    this.setWebsite(origin, firstParty);
+    this.setFirstParty(origin, firstParty);
     this.setWebsite(target, thirdParty);
 
-    if (newThirdParty) {
-      this.updateChild(this.outputWebsite(target, data, origin));
+    if (isNewWebsite && thirdParty['isVisible']) {
+      this.updateChild(this.outputWebsite(target, this._websites[target]));
```
jonathanKingston: The second argument here can just be `thirdParty`

js/store.js
biancadanforth: A relic from where I took it from. Fixed this.

js/store.js
```diff
@@ -112,51 +216,139 @@ const store = {
...
+
     this.setWebsite(hostname, data);
+
+    if (isNewWebsite) {
+      this.updateChild(this.outputWebsite(hostname, this._websites[hostname]));
+    }
   },
 
+  // @todo use data from capture to update UI/filter graph
+  // eslint-disable-next-line no-unused-vars
```
biancadanforth: The way `setThirdParty` is set up right now, I don't actually use the request data, just the `origin` and `target` hostname strings. As a reminder, the request data received from capture is:
```javascript
const data = {
        document: documentUrl.hostname,
        target: targetUrl.hostname,
        origin: originUrl.hostname,
        requestTime: response.timeStamp
      };
```
This data can be merged into the `thirdParty` object, but these values would be overwritten as soon as another unwhitelisted site makes a request to this third party. I figured this would be addressed when we refactor storage and complete #65 . What do you think should be done in the interim?

js/store.js
biancadanforth: Yes; a relic of where I moved it from. Fixed. Thank you.

js/store.js
biancadanforth: No third party is owned by more than one entity (at least for the current Disconnect Entity List, which has 1800+ unique third party sites; I checked this.). Conceptually, the idea that a given third party is owned by more than one entity doesn't make sense to me; though I guess it is possible that the entity list could mistakenly create this situation, and we don't want things to break as a result...

js/store.js
biancadanforth: Agreed; fixed this. Thanks. So many things like that to keep in mind.

js/store.js
biancadanforth: I've updated it to this:

```javascript
const whiteListedFirstParty = thirdParty['firstPartyHostnames']
          .filter((firstPartyHostname) => {
            return !(this._websites[firstPartyHostname]['thirdPartyHostnames']
              .includes(target));
          });
        whiteListedFirstParty.forEach((firstPartyHostname) => {
          this._websites[firstPartyHostname]['thirdPartyHostnames']
            .push(target);
        });
```

js/store.js
biancadanforth: Okay; this is handled with your refactor suggestion (after I debug it). :) In general, passing states between functions is a bad idea, unless you use a state manager (we won't discuss the framework that start with an R that does this).

js/store.js
biancadanforth: I can see for sure that we need to change a few things in your code:
1. Add the code below before the nested block of `if` statements in `setThirdParty`; since the first party won't have this key initialized for the very first third party request it makes.
```javascript
    if (!('thirdPartyHostnames' in firstParty)) {
      firstParty['thirdPartyHostnames'] = [];
    }
```
2. You also never initialize `thirdParty['isVisible'] to false if it is whitelisted. I added an `else` statement for this case.
3. You call `getWebsite` for `firstParty` inside `setThirdParty`, then you call it again in `addFirstPartyRelationship`. Since we're accessing `firstParty['thirdPartyHostnames']` in the former as the top level `if` statement (in the set of nested `ifs`), I have passed `firstParty` into the latter instead of re-getting it there and repeating the same logic to build it up in the helper method. When I remove the redundancy in `addFirstPartyRelationship`, that method only ends up being two lines. To me,  seems like it only saves one line of code; but I can see how it would be useful for clarity.
```javascript
  addFirstPartyRelationship(firstParty, firstPartyHostname, thirdPartyHostname) {
    firstParty['thirdPartyHostnames'].push(target);
    this.setFirstParty(firstPartyHostname, firstParty);
  }
```
4. Also the logic of the nested `if` statements does not account for shared third party sites that are NOT whitelisted for any third party:
Example: visit en.gravatar.com then tylersavin.com. Both link to:
* s0.wp.com
* p.typekit.net
* www.google-analytics.com
* s.gravatar.com (owned by en.gravatar.com)
Because the first 3 third parties are already in firstParty['thirdPartyHostnames'] from visiting en.gravatar.com, their links to this new first party tylersavin.com are not updated.

To fix this, I have created a `shouldUpdate` boolean, which is a combination of three other boolean values:
```javascript
// should update if:
    //   - there is a new third party, OR
    //   - there is a new link between an old third party and a new first party
    //   AND
    //   - the third party is visible
    const shouldUpdate = (isNewThirdParty || isNewFirstParty)
      && thirdParty['isVisible'];

    if (shouldUpdate) {
      this.updateChild(this.outputWebsite(target, thirdParty));
    }
```

jonathanKingston: 1. I intentionally wasn't setting the 1st party in this function for clarity, instead just presence check it or we should think about initing it elsewhere.

Either:
```
this.getFirstParty(origin); // this in tern will call get website and initialise it to have all the data we care about
```
Or:
```
if (!('thirdPartyHostnames' in  firstParty)
    || !firstParty['thirdPartyHostnames'].includes(target)) { 
```

2. yup 👍 I said it was untested and you were the tester haha

3. this logic is wrong, you are overriding abc.xyz and google.co.uk with the data from mozilla.org when all 3 load googleanalytics.com This is why I explicitly was only getting in this other function. The other first party get is just to check if the link exists between mozilla.org and GA. This is irrelevant mostly to the data we set in the other function.

4. Nah I flat out missed:
```
    if (!('isVisible' in thirdParty)
        || !thirdParty.isVisible) { 
      ...
   } else {
     this.addFirstPartyRelationship(origin, target);
   }
```

js/store.js
jonathanKingston: @biancadanforth google analytics is whitelisted by google.de, google.co.uk abc.xyz what am I missing?

js/store.js
```diff
@@ -273,80 +274,78 @@ const store = {
...
+      isNewFirstParty = true;
+    }
+
+    // add link in first party if it doesn't exist yet
+    // and the third party is visible (i.e. not whitelisted)
+    if (!('thirdPartyHostnames' in firstParty)) {
+      firstParty['thirdPartyHostnames'] = [];
+    }
+    if (!firstParty['thirdPartyHostnames'].includes(target)) {
+      if (!('isVisible' in thirdParty) || !thirdParty['isVisible']) {
```
jonathanKingston: Probs agrument to have this as a `isVisible` or `isWebsiteVisible` fn call.

js/store.js
jonathanKingston: This seems fine for now, maybe add a comment as we want to do something more smart in merging these values. However this is a known bug for first party as well.

js/store.js
biancadanforth: Yes, you're right. I misunderstood you.

js/store.js
biancadanforth: 1. Okay I moved that stuff back to the `addFirstPartyLink` method.
2. Ok. :)
3. I moved this per #1. But how is this logic wrong? We're only handling 1 request at a time. Are you referring to the request data (`data`) passed in from capture being overwritten? There's [another comment](https://github.com/electrolyfish/lightbeam-we/pull/109/files/a9fee22d02a22e32076518c0c6dda8622185c30e#diff-33ff4b4d14285f8f3d55906c0416c235R281) about that part.
4. Ok.

jonathanKingston: 3. We never store in each first party the third party connection such that the diagram does not draw a third party that we will not be telling them about.

However when we make this node visible we need to update all these connections to link back up to the third party for when the user refreshes the lightbeam page and also just for consistency with how we store it elsewhere.

This first party we are handling right now is the one that activates the visibility, however we also have a tonne of nodes that don't link to the third party that need to be updated too.

jonathanKingston: Also test to make sure my fixes solve all your issues. Assume it's psudo code as I haven't even run it. :) thanks

biancadanforth: Yes I have tested my implementation of your pseudocode :)

jonathanKingston: 4. doesn't look like it was addressed? Essentially that first party connection is never connected for visible nodes that are not new.

js/store.js
```diff
@@ -112,49 +217,137 @@ const store = {
...
+        // an existing visible third party links to a new first party
+        thirdParty['firstPartyHostnames'].forEach((firstPartyHostname) => {
+          this.addFirstPartyLink(firstPartyHostname, target);
+        });
+      }
+    }
+
+    // merge data with thirdParty website object
+    for (const key in data) {
+      thirdParty[key] = data[key];
```
jonathanKingston: This should be removed and `setWebsite` should get data as a third argument

js/store.js
```diff
@@ -112,49 +217,137 @@ const store = {
...
+        if (!(this.onWhiteList(origin, target))) {
+          // show third party
+          thirdParty['isVisible'] = true;
+          isNewThirdParty = true;
+        } else {
+          // hide third party
+          thirdParty['isVisible'] = false;
+        }
+      }
+      if (thirdParty['isVisible'] && isNewFirstParty) {
```
jonathanKingston: Nit: consistent new line on second argument?
```
if (thirdParty['isVisible']
    && isNewFirstParty) {
```

Shouldn't this be presence checking `isVisible` like elsewhere?

This also could be turned to a method:
```
if (this.isWebsiteVisible(thirdParty)
    && isNewFirstParty) {
```
Where `isWebsiteVisible` checks both the key exists and is true.

js/store.js
```diff
@@ -112,49 +217,137 @@ const store = {
     }
 
     this._write(websites);
+  },
+
+  isNewWebsite(hostname) {
+
```
jonathanKingston: Nit: ditch this new line

js/store.js
biancadanforth: Can you give me a specific example of what sites this is happening for?

js/store.js
```diff
@@ -112,49 +217,137 @@ const store = {
...
+        } else {
+          // hide third party
+          thirdParty['isVisible'] = false;
+        }
+      }
+      if (thirdParty['isVisible'] && isNewFirstParty) {
+        // an existing visible third party links to a new first party
+        thirdParty['firstPartyHostnames'].forEach((firstPartyHostname) => {
+          this.addFirstPartyLink(firstPartyHostname, target);
+        });
```
jonathanKingston: This code:
```
        thirdParty['firstPartyHostnames'].forEach((firstPartyHostname) => {
          this.addFirstPartyLink(firstPartyHostname, target);
        });
```

Should go after `// show third party` as it's about the third party connecting all it's links with first parties.

jonathanKingston: However also missing is the:
```
this.addFirstPartyRelationship(origin, target);
```
This can happen where this code is or somewhere inside the following if statement:
```
    if (!('thirdPartyHostnames' in  firstParty)
|| !firstParty['thirdPartyHostnames'].includes(target)) {
```

The for loop could also ignore the origin too perhaps but that also makes following the logic complex.

js/store.js
```diff
@@ -112,49 +217,137 @@ const store = {
...
     }
     if (!thirdParty['firstPartyHostnames'].includes(origin)) {
       thirdParty['firstPartyHostnames'].push(origin);
-      newThirdParty = true;
+      isNewFirstParty = true;
+    }
+
+    // add link in first party if it doesn't exist yet
+    // and the third party is visible (i.e. not whitelisted)
+    if (!('thirdPartyHostnames' in  firstParty)
```
jonathanKingston: Nit double space after `in`

js/store.js
jonathanKingston: Also not convinced we need `isNewFirstParty` and instead we actually care more about `isNewThirdParty`, see below where I suggest nesting it in that code instead.

jonathanKingston: For clarity I think the code should be something like this:

```
if (!this.isFirstPartyConnectedToThirdParty(firstParty)) {
  if (!this.isVisibleThirdParty(thirdParty)) {
    if (this.onWhiteList(origin, target)) {
      // Third party starts off as invisible
      thirdParty['isVisible'] = false;
    } else {
      /* Third party becomes visible!
         this broadcasts updates and makes connections to all previous seen first parties */
      thirdParty['isVisible'] = true;
      isNewThirdParty = true;
      thirdParty['firstPartyHostnames'].forEach((firstPartyHostname) => {
        this.addFirstPartyLink(firstPartyHostname, target);
      });
    }   
  }
  if (thirdParty['isVisible']
      && !isNewThirdParty) {
    // an existing visible third party links to a new first party
    this.addFirstPartyLink(origin, target);
  }   
}
```

js/store.js
biancadanforth: Done

js/store.js
biancadanforth: We don't need the presence check in this case, as `thirdParty` will definitely have a `isVisible` key by that point in the code.

We need to call update in the case there's a new third party AND in the case there's an old third party with a new first party. To me, it seems clearer to leave that boolean there, since we call update in that case not because of a new third party, but because of a new first party. Though I understand the boolean bloat that is going on right now here.

Anyway, I implemented it as you suggest (removing `isNewFirstParty`).

js/store.js
biancadanforth: Done

js/store.js
biancadanforth: `this.addFirstPartyRelationship(origin, target);` is not missing. For line length considerations I changed the method name to `this.addFirstPartyLink`. When I iterate through each element in `thirdParty['firstPartyHostnames']`, I add the link for the current `origin` and `target` pair as well.

What I did realize was that I didn't need to do the `forEach` loop for the case of a new first party, old third party.

js/store.js
biancadanforth: Fixed.

js/store.js
```diff
@@ -296,58 +300,64 @@ const store = {
...
-      && thirdParty['isVisible'];
+    this.setWebsite(target, thirdParty, data);
 
     if (shouldUpdate) {
       this.updateChild(this.outputWebsite(target, thirdParty));
     }
   },
 
+  isFirstPartyConnectedToThirdParty(firstParty, thirdPartyHostname) {
+    if (!('thirdPartyHostnames' in  firstParty)
```
jonathanKingston: Nit: double space.

js/store.js
```diff
@@ -296,58 +300,64 @@ const store = {
...
+
   addFirstPartyLink(firstPartyHostname, thirdPartyHostname) {
     const firstParty = this.getWebsite(firstPartyHostname);
     if (!('thirdPartyHostnames' in firstParty)) {
       firstParty['thirdPartyHostnames'] = [];
     }
-    firstParty['thirdPartyHostnames'].push(thirdPartyHostname);
-    this.setFirstParty(firstPartyHostname, firstParty);
+    if (!this
+      .isFirstPartyConnectedToThirdParty(firstParty, thirdPartyHostname)) {
```
jonathanKingston: Nit wrapping here is pretty hard to read.

js/store.js
jonathanKingston: Yup I updated my other example to look like this so this seems correct. I think I'm going to propose a refactor at some point later on as I have a follow up idea.

js/store.js
```diff
@@ -205,15 +205,19 @@ const store = {
...
       websites[hostname] = {};
     }
 
-    for (const key in data) {
-      websites[hostname][key] = data[key];
+    for (const key in partyData) {
+      websites[hostname][key] = partyData[key];
+    }
+
+    for (const key in requestData) {
```
jonathanKingston: I actually preferred it where it was, I was being tired sorry about that.

jonathanKingston: Feel free to leave this where it is for now though.

js/store.js
```diff
@@ -296,58 +300,64 @@ const store = {
...
         }
       }
-      if (thirdParty['isVisible'] && isNewFirstParty) {
+      if (this.isVisibleThirdParty(thirdParty) && !isNewThirdParty) {
         // an existing visible third party links to a new first party
-        thirdParty['firstPartyHostnames'].forEach((firstPartyHostname) => {
-          this.addFirstPartyLink(firstPartyHostname, target);
-        });
+        this.addFirstPartyLink(origin, target);
+        shouldUpdate = true;
```
jonathanKingston: Again we might want to readress this later but yeah this logic seems correct now thanks.

js/store.js
```diff
@@ -296,58 +300,64 @@ const store = {
...
-        } else {
+    if (!this.isFirstPartyConnectedToThirdParty(firstParty, target)) {
+      if (!this.isVisibleThirdParty(thirdParty)) {
+        if (this.onWhiteList(origin, target)) {
           // hide third party
           thirdParty['isVisible'] = false;
+        } else {
+          // show third party and all its connections
+          thirdParty['isVisible'] = true;
+          isNewThirdParty = true;
```
jonathanKingston: Nit: this var now represents `becameVisible` it would be clearer for future Jonathan and Bianca to call it that :) (being that this algo is horrible hah)

jonathanKingston: By horrible I mean more complex that it seems.

js/store.js
biancadanforth: Fixed.

js/store.js
biancadanforth: Okay yeah this was due to the 80 character line length limit in eslint. I shortened the name of this method slightly to `isFirstPartyLinkedToThirdParty`. :)

js/store.js
biancadanforth: Ok I moved it back.

js/store.js
biancadanforth: Noted

js/store.js
biancadanforth: Well, this `else` statement actually handles two cases at the same time: the case that a previously hidden third party becomes visible, and the case that a new third party that the first party has not linked to before and is not on the whitelist is present. Think about it as "new to the visualization", not "new to storage".

biancadanforth: What if I add a helpful comment like:
`// show third party; it either became visible or is brand new`

jonathanKingston:
Congrats 👍 - I know this wasn't easy.
I will file any follow ups we might need.

Issue #107, opened 2017-07-18
Add high DPI scaling to canvas


jonathanKingston:
The following made the graph look a lot more like how the SVG did on a high DPI screen.
```
diff --git a/css/style.css b/css/style.css
index a15a121..f5ae101 100644
--- a/css/style.css
+++ b/css/style.css
@@ -290,7 +290,8 @@ input[type="checkbox"] {
   color: var(--secondary-text-color);
 }
 
-#visualization {
+#visualization,
+#visualization canvas {
   width: 500px;
   height: 400px;
 }
diff --git a/js/viz.js b/js/viz.js
index 5f931d3..bec04b3 100644
--- a/js/viz.js
+++ b/js/viz.js
@@ -54,9 +54,11 @@ const viz = {
   createCanvas(width, height) {
     const base = d3.select('#visualization');
     const canvas = base.append('canvas');
-    canvas.attr('width', width);
-    canvas.attr('height', height);
+    const scaling = 4;
+    canvas.attr('width', width * scaling);
+    canvas.attr('height', height * scaling);
     const context = canvas.node().getContext('2d');
+    context.scale(scaling, scaling);
 
     return context;
   },
```

However what I think we should be doing is instead of hardcoding the width and hight in CSS is using relative sizing and changing the context size and DPI based on that. This will require redrawing the graph when the screen changes size. Either way you need to make sure that the physical size is never overriden by those canvas sizes.

PR #112, opened 2017-07-27
Canvas DPI and UI responsiveness

princiya:
Issue #107 

**Before:**

screen shot 2017-07-27 at 11 47 45

**After:**

screen shot 2017-07-27 at 11 47 23

**To test Lightbeam on retina display:** [Link](https://stackoverflow.com/questions/12243549/how-to-test-a-webpage-meant-for-retina-display)

screen shot 2017-07-27 at 11 52 32

screen shot 2017-07-27 at 11 53 08

### DIFF
--- a/css/style.css
+++ b/css/style.css
@@ -205,6 +205,7 @@ main {
   display: grid;
   grid-template-rows: 10% 90%;
   background-color: var(--primary-color);
+  height: 100vh;
 }
 
 .top-bar {
@@ -299,6 +300,7 @@ input[type="checkbox"] {
   position: absolute;
   right: 0px;
   top: 30px;
+  display: none;
 }
 
 .info-panel {
--- a/css/style.css
+++ b/css/style.css
@@ -203,9 +203,8 @@ aside h2 {
 
 main {
   display: grid;
-  grid-template-rows: 10% 90%;
+  grid-template-rows: 10% 60% 30%;
   background-color: var(--primary-color);
-  height: 100vh;
 }
 
 .top-bar {
@@ -264,24 +263,23 @@ input[type="checkbox"] {
 
 .vis {
   display: grid;
-  grid-template-rows: 75% 25%;
-  position: relative;
+  grid-template-rows: 10% 90%;
+  grid-template-columns: 1fr 40px;
 }
 
-.vis-panel {
+.vis-header {
   padding: 30px;
-  padding-bottom: 0;
 }
 
-.vis-panel > h1 {
+.vis-header > h1 {
   color: var(--primary-text-color);
 }
 
-.vis-panel > h2 {
+.vis-header > h2 {
   color: var(--secondary-text-color);
 }
 
-#visualization {
+.vis-content {
   width: 100%;
   height: 100%;
   position: relative;
@@ -296,11 +294,12 @@ input[type="checkbox"] {
   padding: 4px;
 }
 
+.vis-controls {
+  grid-row-end: span 3;
+}
+
 .info-panel-controls {
-  position: absolute;
-  right: 0px;
-  top: 30px;
-  display: none;
+
 }
 
 .info-panel {
@@ -334,11 +333,10 @@ input[type="checkbox"] {
 /* ----------- UI - vis Controls ----------- */
 
 footer {
-  grid-row-start: 2;
   display: grid;
   grid-template-columns: repeat(4, 1fr) 150px;
+  grid-row-start: 3;
   padding: 30px;
-  padding-top: 0;
 }
 
 footer h2 {
--- a/css/style.css
+++ b/css/style.css
@@ -263,36 +263,27 @@ input[type="checkbox"] {
 /* ----------- UI - Graph ----------- */
 
 .vis {
-  display: flex;
-  flex-direction: column;
-  justify-content: space-between;
+  width: 100%;
+  height: 90%;
   padding: 30px;
+  position: relative;
 }
 
-.vis > div {
-  display: flex;
-  align-items: flex-start;
-  justify-content: space-between;
-  height: 100%;
+.vis-panel {
+  height: 75%;
 }
 
-.vis > div > div {
-  display: flex;
-  flex-direction: column;
-  height: 100%;
-}
-
-.vis h1 {
+.vis-panel > h1 {
   color: var(--primary-text-color);
 }
 
-.vis h2 {
+.vis-panel > h2 {
   color: var(--secondary-text-color);
 }
 
 #visualization {
-  width: 500px;
-  height: 400px;
+  width: 100%;
+  height: 100%;
   position: relative;
 }
 
@@ -306,10 +297,9 @@ input[type="checkbox"] {
 }
 
 .info-panel-controls {
-  display: flex;
-  flex-direction: column;
-  margin-right: -30px;
-  margin-top: -30px;
+  position: absolute;
+  right: 30px;
+  top: 30px;
 }
 
 .info-panel {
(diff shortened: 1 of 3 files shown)


js/viz.js
```diff
@@ -60,9 +60,14 @@ const viz = {
...
-    canvas.attr('width', width);
-    canvas.attr('height', height);
     const context = canvas.node().getContext('2d');
+    const scale = window.devicePixelRatio || 1;
+
+    canvas.attr('width', width * scale);
+    canvas.attr('height', height * scale);
+    canvas.style('width', `${width}px`);
+    canvas.style('height', `${height}px`);
+    context.scale(scale, scale);
```
princiya: Following is the approach used:

Creating the Canvas element in the scaled up resolution and drawing in the context which is scaled accordingly.

The following two lines ensures that overall graph width, height is never overridden by the scaled up values.
```
canvas.style('width', `${width}px`);
canvas.style('height', `${height}px`);
```

princiya:
@jonathanKingston This PR doesn't redraw the graph when the screen size changes. Shall we tackle that part when we make the UI responsive?

princiya:
@biancadanforth not sure if I have messed up with the styles in this PR in an attempt to clean-up some unused divs. Please test and let me know :)

js/viz.js
jonathanKingston: Please use native DOM methods instead of whatever d3 exposes to stay consistent with all other code we have.

```
canvas.setAttribute('width', width * scale);
```
Or

```
canvas.width = width * scale;
```

and:

```
canvas.style.width = `${width}px`;
```

jonathanKingston:
@princiya follow up is fine. I added one comment, let me know when that is switched over so I can test.
It should prevent us being coupled to d3 if we need that etc.

Did you do any experimenting with making the scale factor bigger than what the user has set? For example when I tried I ended up with a much neater looking diagram again with 3 or 4 as the scale factor. If you end up using a bigger scale factor above `devicePixelRatio` you may need to scale up things also so:
```
 scalingFactor: 2,
 circleRadius: 5 * this.scalingFactor,
...
 const scale = (window.devicePixelRatio || 1) * this.scalingFactor;
...
```

Give that a whirl if you can, it might end up breaking everything so could also be a follow up.

biancadanforth:
@princiya: I don't see anything major broken in the UI; you already mentioned what I was going to point out about the visualization not resizing on window resize; I vote for a follow-up issue for that. Resolution looks great.

princiya:
@biancadanforth thanks for testing.

@jonathanKingston added changes.

jonathanKingston:
Looks like the buttons on the right are positioned wrong now not sure if this was part of this change.

jonathanKingston:
@princiya also looks like the bottom of the graph on a small window overlaps onto the controls too.

Sometimes not visible because of how the nodes fall however try this:
```
function drawything(i) { if (i==0) {return;} viz.simulate(); viz.drawOnCanvas(); setTimeout(() => {drawything(i-1);}, 2); } drawything(1000);
```

The above code also points out the issue about the initial render of nodes when the graph is started too.

princiya:
@jonathanKingston thanks. yes, I thought some styles might have been messed up. looking into :)

biancadanforth:
Sorry that my cursory look missed some obvious things!

I neglected to try it with some busy website graphs like with fastcompany.com followed by youtube.com. I noticed this when the window size is more narrow:

screen shot 2017-07-28 at 10 04 31 am

On inspection, looks like the `.vis` element is going outside the bounds of its parent element, `main`, among some other things.

I tried out some changes, and it looks like reverting the changes to `.vis` and `.vis-panel` styles , adding `position: relative` to `.vis`, and removing `min-width` from `main` fixes this.

Also, the `.info-panel-controls` originally used `position: absolute` as you have here, but in the initial UI PR (which I of course, overly rebased and now no longer has jkt's original comment), it was pointed out that we should avoid absolute positioning.
Perhaps @jonathanKingston can elaborate more on why? and should we leave the info panel as is or not? Those extra divs that you removed a while back were being used to use flexbox there instead. [Here's the commit](https://github.com/MozillaFoundation/mozfest-program-2017/issues) I am referencing.

princiya:
@biancadanforth thanks for the updates. I don't intend to do any UI related changes here, but I don't want things to be broken too. So I will see what's a quick fix and then we can go back and polish the UI.

biancadanforth:
@princiya The broken part that I am referring to are due to changes made in this PR; see my previous comment for which changes I'm talking about.

css/style.css
```diff
@@ -263,36 +263,27 @@ input[type="checkbox"] {
...
-}
-
-.vis > div {
-  display: flex;
-  align-items: flex-start;
-  justify-content: space-between;
-  height: 100%;
+  display: grid;
+  grid-template-rows: 75% 25%;
+  position: relative;
```
biancadanforth: Per my [comment](https://github.com/electrolyfish/lightbeam-we/pull/112#issuecomment-318718783) on the PR, reverting `vis` is part 1 of 3 of the fix.

css/style.css
```diff
@@ -263,36 +263,27 @@ input[type="checkbox"] {
...
 }
 
-.vis > div > div {
-  display: flex;
-  flex-direction: column;
-  height: 100%;
+.vis-panel {
+  padding: 30px;
+  padding-bottom: 0;
```
biancadanforth: Per my [comment](https://github.com/electrolyfish/lightbeam-we/pull/112#issuecomment-318718783) on the PR, reverting `.vis-panel` is part 2 of 3 of the fix. The divs are not exactly the same here since the HTML was restructured, but it's a combination of the two blocks that were removed:

```css
.vis-panel {
  display: flex;
  flex-direction: column;
  height: 100%;
}
```

css/style.css
```diff
@@ -346,11 +336,14 @@ footer {
   grid-row-start: 2;
   display: grid;
   grid-template-columns: repeat(4, 1fr) 150px;
+  padding: 30px;
+  padding-top: 0;
```
biancadanforth: An aside from this [comment](https://github.com/electrolyfish/lightbeam-we/pull/112#issuecomment-318718783) on the PR, since `footer`'s container element has padding, reverting this (aka removing these two padding lines) would keep the whitespace consistent as part of the fix.

css/style.css
biancadanforth: Part 3 of 3 is to remove `min-width: 780px` from `main` (circa line 202 in the CSS). You didn't edit that line in this PR; it was added at some point earlier (may have been me, honestly!).

princiya:
work in progress... few updates coming! :)

princiya:
Graph resize:

Resize

css/style.css
```diff
@@ -263,36 +266,24 @@ input[type="checkbox"] {
...
-  padding: 30px;
-}
-
-.vis > div {
-  display: flex;
-  align-items: flex-start;
-  justify-content: space-between;
-  height: 100%;
+  display: grid;
+  grid-template-rows: 1fr 2fr;
```
princiya: I love this rule! So neat when it comes to responsiveness!

princiya:
Follow-up issue: https://github.com/electrolyfish/lightbeam-we/issues/114

js/viz.js
```diff
@@ -143,19 +156,48 @@ const viz = {
...
       }
     });
+  },
+
+  addWindowResize() {
+    window.addEventListener('resize', () => {
+      clearTimeout(this.resizeTimer);
+      this.resizeTimer = setTimeout(() => {
+        this.onResize();
+      }, 250);
```
princiya: debouncing the resize event.

princiya:
The footer can still be improved. And off-course other nit-picks when it comes to UI details :)

biancadanforth:
Cool; nice initial work on responsiveness; like you say there's a lot of small details for it. Yes CSS Grid is pretty cool! It's worth making responsiveness its own issue if it isn't already there.

PR #115, opened 2017-08-03
Initial swap of localStorage for IndexedDB via Dexie

### LINKED ISSUE #13: Choose client-side data storage
In our MVP web extension, we use the _storage_ API, but it may not suit our needs long term (e.g. for advanced filtering). We should figure out what our storage requirements are and what the most appropriate option will be (other options: indexedDB, SQLite...).

biancadanforth:
WIP; Fixes #13  when complete.

### DIFF
--- a/js/capture.js
+++ b/js/capture.js
@@ -46,7 +46,8 @@ const capture = {
         document: documentUrl.hostname,
         target: targetUrl.hostname,
         origin: originUrl.hostname,
-        requestTime: response.timeStamp
+        requestTime: response.timeStamp,
+        firstParty: false
       };
       store.setThirdParty(
         documentUrl.hostname,
@@ -60,7 +61,10 @@ const capture = {
   sendFirstParty(tabId, changeInfo, tab) {
     const documentUrl = new URL(tab.url);
     if (tab.status === 'complete' && this.shouldStore(tab)) {
-      const data = { faviconUrl: tab.favIconUrl };
+      const data = {
+        faviconUrl: tab.favIconUrl,
+        firstParty: true
+      };
       store.setFirstParty(documentUrl.hostname, data);
     }
   }
--- a/js/capture.js
+++ b/js/capture.js
@@ -3,6 +3,7 @@
 * third-party requests to storage.
 */
 const capture = {
+
   init() {
     this.addListeners();
   },
@@ -35,11 +36,17 @@ const capture = {
       });
   },
 
+  // Process each HTTP request or tab page load in order,
+  // so that async reads/writes to IndexedDB
+  // (via sendFirstParty and sendThirdParty) won't miss data
+  // The 'ignore' boolean ensures processNextEvent is only
+  // executed when the previous call to processNextEvent
+  // has completed.
   async processNextEvent(ignore = false) {
     if (this.processingQueue && !ignore) {
       return;
     }
-    if (this.queue.length > 1) {
+    if (this.queue.length >= 1) {
       const nextEvent = this.queue.shift();
       this.processingQueue = true;
       switch (nextEvent.type) {
@@ -54,7 +61,9 @@ const capture = {
           await this.sendThirdParty(nextEvent.data);
           break;
         default:
-          break;
+          throw new Error(
+            'An event must be of type sendFirstParty or sendThirdParty.'
+          );
       }
       this.processNextEvent(true);
     } else {
@@ -65,11 +74,10 @@ const capture = {
   // Returns true if the request should be stored, otherwise false.
   shouldStore(tab) {
     const documentUrl = new URL(tab.url);
-    // ignore about:*, moz-extension:* & non-visible tabs (like dev tools)
+    // ignore about:*, moz-extension:*
     // also ignore private browsing tabs
     if (documentUrl.protocol !== 'about:'
       && documentUrl.protocol !== 'moz-extension:'
-      && tab.id !== browser.tabs.TAB_ID_NONE
       && !tab.incognito) {
       return true;
     }
@@ -78,14 +86,19 @@ const capture = {
 
   // capture third party requests
   async sendThirdParty(response) {
+    // browser.tabs.get throws an error for non-visible tabs (like DevTools)
+    if (response.tabId === browser.tabs.TAB_ID_NONE) {
+      return;
+    }
     const tab = await browser.tabs.get(response.tabId);
     const documentUrl = new URL(tab.url);
     const targetUrl = new URL(response.url);
+    // @todo figure out why Web Extensions sometimes gives
+    // undefined for response.originUrl
+    const originUrl = !response.originUrl ? '' : new URL(response.originUrl);
 
     if (targetUrl.hostname !== documentUrl.hostname
       && this.shouldStore(tab)) {
-      // @todo figure out why Web Extensions sometimes gives us undefined URLs
-      const originUrl = new URL(response.originUrl);
       const data = {
         document: documentUrl.hostname,
         target: targetUrl.hostname,
--- a/js/capture.js
+++ b/js/capture.js
@@ -26,9 +26,9 @@ const capture = {
         const eventDetails = {
           type: 'sendFirstParty',
           data: {
-            tabId: tabId,
-            changeInfo: changeInfo,
-            tab: tab
+            tabId,
+            changeInfo,
+            tab
           }
         };
         this.queue.push(eventDetails);
@@ -72,13 +72,28 @@ const capture = {
   },
 
   // Returns true if the request should be stored, otherwise false.
-  shouldStore(tab) {
-    const documentUrl = new URL(tab.url);
+  // info could be a tab (from setFirstParty) or a
+  // response (from setThirdParty) object
+  async shouldStore(info) {
+    const tabId = info.id || info.tabId;
+    const visibleTab = tabId !== browser.tabs.TAB_ID_NONE;
+    let tab, documentUrl, privateBrowsing;
+    if (visibleTab) {
+      tab = await browser.tabs.get(tabId);
+      documentUrl = new URL(tab.url);
+      privateBrowsing = tab.incognito;
+    } else {
+      // browser.tabs.get throws an error for nonvisible tabs (tabId = -1)
+      // but some non-visible tabs can make third party requests,
+      // ex: Service Workers
+      documentUrl = new URL(info.documentUrl);
+      privateBrowsing = false;
+    }
     // ignore about:*, moz-extension:*
     // also ignore private browsing tabs
     if (documentUrl.protocol !== 'about:'
       && documentUrl.protocol !== 'moz-extension:'
-      && !tab.incognito) {
+      && !privateBrowsing) {
       return true;
     }
     return false;
@@ -86,19 +101,20 @@ const capture = {
 
   // capture third party requests
   async sendThirdParty(response) {
-    // browser.tabs.get throws an error for non-visible tabs (like DevTools)
-    if (response.tabId === browser.tabs.TAB_ID_NONE) {
+    if (!response.documentUrl) {
+      // documentUrl is undefined for the first request from the browser to the
+      // first party site
       return;
     }
-    const tab = await browser.tabs.get(response.tabId);
-    const documentUrl = new URL(tab.url);
-    const targetUrl = new URL(response.url);
+
     // @todo figure out why Web Extensions sometimes gives
     // undefined for response.originUrl
-    const originUrl = !response.originUrl ? '' : new URL(response.originUrl);
+    const originUrl = response.originUrl ? new URL(response.originUrl) : '';
+    const documentUrl = new URL(response.documentUrl);
+    const targetUrl = new URL(response.url);
 
     if (targetUrl.hostname !== documentUrl.hostname
-      && this.shouldStore(tab)) {
+      && await this.shouldStore(response)) {
       const data = {
         document: documentUrl.hostname,
         target: targetUrl.hostname,
@@ -117,7 +133,7 @@ const capture = {
   // capture first party requests
   async sendFirstParty(tabId, changeInfo, tab) {
     const documentUrl = new URL(tab.url);
-    if (tab.status === 'complete' && this.shouldStore(tab)) {
+    if (tab.status === 'complete' && await this.shouldStore(tab)) {
       const data = {
         faviconUrl: tab.favIconUrl,
         firstParty: true
--- a/js/capture.js
+++ b/js/capture.js
@@ -9,14 +9,57 @@ const capture = {
 
   addListeners() {
     // listen for each HTTP response
-    browser.webRequest.onResponseStarted.addListener(
-      (response) => this.sendThirdParty(response),
+    this.queue = [];
+    browser.webRequest.onResponseStarted.addListener((response) => {
+      const eventDetails = {
+        type: 'sendThirdParty',
+        data: response
+      };
+      this.queue.push(eventDetails);
+      this.processNextEvent();
+    },
       {urls: ['<all_urls>']});
     // listen for tab updates
     browser.tabs.onUpdated.addListener(
       (tabId, changeInfo, tab) => {
-        this.sendFirstParty(tabId, changeInfo, tab);
+        const eventDetails = {
+          type: 'sendFirstParty',
+          data: {
+            tabId: tabId,
+            changeInfo: changeInfo,
+            tab: tab
+          }
+        };
+        this.queue.push(eventDetails);
+        this.processNextEvent();
       });
+  },
+
+  async processNextEvent(ignore = false) {
+    if (this.processingQueue && !ignore) {
+      return;
+    }
+    if (this.queue.length > 1) {
+      const nextEvent = this.queue.shift();
+      this.processingQueue = true;
+      switch (nextEvent.type) {
+        case 'sendFirstParty':
+          await this.sendFirstParty(
+            nextEvent.data.tabId,
+            nextEvent.data.changeInfo,
+            nextEvent.data.tab
+          );
+          break;
+        case 'sendThirdParty':
+          await this.sendThirdParty(nextEvent.data);
+          break;
+        default:
+          break;
+      }
+      this.processNextEvent(true);
+    } else {
+      this.processingQueue = false;
+    }
   },
 
   // Returns true if the request should be stored, otherwise false.
@@ -38,10 +81,11 @@ const capture = {
     const tab = await browser.tabs.get(response.tabId);
     const documentUrl = new URL(tab.url);
     const targetUrl = new URL(response.url);
-    const originUrl = new URL(response.originUrl);
 
     if (targetUrl.hostname !== documentUrl.hostname
       && this.shouldStore(tab)) {
+      // @todo figure out why Web Extensions sometimes gives us undefined URLs
+      const originUrl = new URL(response.originUrl);
       const data = {
         document: documentUrl.hostname,
         target: targetUrl.hostname,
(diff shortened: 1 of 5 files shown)


js/store.js
```diff
@@ -349,21 +354,20 @@ const store = {
...
-  addFirstPartyLink(firstPartyHostname, thirdPartyHostname) {
+  async addFirstPartyLink(firstPartyHostname, thirdPartyHostname) {
     const firstParty = this.getWebsite(firstPartyHostname);
     if (!('thirdPartyHostnames' in firstParty)) {
       firstParty['thirdPartyHostnames'] = [];
     }
     if (!this.isFirstPartyLinkedToThirdParty(firstParty, thirdPartyHostname)) {
       firstParty['thirdPartyHostnames'].push(thirdPartyHostname);
-      this.setFirstParty(firstPartyHostname, firstParty);
+      await this.setFirstParty(firstPartyHostname, firstParty);
```
jonathanKingston: Can you make this an array of promises then Promise.all it?

js/store.js
```diff
@@ -192,35 +205,26 @@ const store = {
...
-    const firstParty = this.getWebsite(hostname);
-    if ('thirdPartyHostnames' in firstParty) {
-      return firstParty.thirdPartyHostnames;
-    }
-
-    return null;
-  },
-
   setWebsite(hostname, data) {
     const websites = clone(this._websites);
```
jonathanKingston: This object cache worries me, we need to look into where data needs to be awaited. if that is in the setting process fine. Perhaps all set methods need to be async (we don't always need to await the response either).

js/store.js
```diff
@@ -308,9 +312,10 @@ const store = {
           // show third party; it either became visible or is brand new
           thirdParty['isVisible'] = true;
           isNewThirdParty = true;
-          thirdParty['firstPartyHostnames'].forEach((firstPartyHostname) => {
-            this.addFirstPartyLink(firstPartyHostname, target);
-          });
+          thirdParty['firstPartyHostnames']
+            .forEach((firstPartyHostname) => {
+              this.addFirstPartyLink(firstPartyHostname, target);
```
jonathanKingston: This might need to be awaited here perhaps and the whole function be async.

js/store.js
```diff
@@ -151,13 +163,14 @@ const store = {
...
-      const website = this._websites[hostname];
+    for (const website of websites) {
       // if it's a visible third party or a first party
       if (website['isVisible'] || !('isVisible' in website)) {
-        output[hostname] = this.outputWebsite(hostname, website);
+        output[website.hostname]
+          = this.outputWebsite(website.hostname, website);
       }
     }
     return Promise.resolve(output);
```
jonathanKingston: As you made this async, returning just `output` will be fine here.

js/store.js
```diff
@@ -151,13 +163,14 @@ const store = {
...
+  async getAll() {
+    const websites = await this.db.websites.toArray();
     const output = {};
-    for (const hostname in this._websites) {
-      const website = this._websites[hostname];
+    for (const website of websites) {
       // if it's a visible third party or a first party
       if (website['isVisible'] || !('isVisible' in website)) {
-        output[hostname] = this.outputWebsite(hostname, website);
+        output[website.hostname]
```
jonathanKingston: Could this be a filter instead? I'm not a massive fan of the !isVisible to check if it's a FP. Perhaps we could improve that later (raise a separate bug please)?

js/store.js
biancadanforth: I'm not sure what you mean here... What would be the array of Promises? You mean capturing `this.setFirstParty` into an array that persists across calls to `addFirstPartyLink`? Could you provide some pseudocode to help illustrate?

js/store.js
biancadanforth: You mean make the arrow function inside `forEach` async to add an await for `addFirstPartyLink`? This would prevent adding the next link until the previous one had completed and done a write?

js/store.js
biancadanforth: Are you referring to the clone of `this._websites` or the existence of `this._websites` itself? This method is based on the original method in terms of the clone. I could remove `websites` and modify `this._websites` directly -- otherwise are you suggesting to make setWebsite async and to await `this._write`?

js/store.js
biancadanforth: By filter, do you mean adding another key to each website object (`firstParty: true`)? This is added in the data to send to the visualization, but it's not kept in storage currently.

js/store.js
jonathanKingston: The existence of `this._websites` I think we can kill it now. I don't think anything should await unless it needs to.

js/store.js
jonathanKingston: I meant for the other use of ` this.addFirstPartyLink(firstPartyHostname, target);` sorry.

js/store.js
jonathanKingston: Change this to an array of promises and await that as mentioned on the wrong line :D

js/store.js
jonathanKingston: By filter I mean query using dexie for isVisible.

The follow up can use a first party key in storage yeah.

js/store.js
biancadanforth: @jonathanKingston Can you try building this PR and looking at the console? `store.addFirstPartyLink` gets called a bunch of times before it finishes writing the `firstParty` object to the database from the very first call... What is happening here, and how can I fix it?

jonathanKingston: @biancadanforth ugh Github is killing me with hiding comments.

Er well I would fix the other issues first, however if this is an issue you will need to make sure that setFirstParty blocks the thread of the capture code otherwise this will happen. The alternative if we start to worry about this blocking is to write it to another table.

So relationships could be stored in a different table reducing the risk of blocking to the same row. However this is more work that we could leave to be a perf enhancements.

biancadanforth: When you say...

> make sure that setFirstParty blocks the thread of the capture code otherwise this will happen

What would that implementation look like? This is not the same as adding or removing `await` per our other discussion [here](https://github.com/biancadanforth/lightbeam-we/commit/5abf415cc1042dc08373d8db9b44c04765c975f5#commitcomment-23517430), is it?

jonathanKingston: It means await in capture for a response.

biancadanforth: Just putting `await` infront of `store.setThirdParty` didn't make a difference for me. Why not?

```javascript
await store.setThirdParty(
        documentUrl.hostname,
        targetUrl.hostname,
        data
      );
```

biancadanforth:
@jonathanKingston Thanks again for helping clarify things for me this afternoon. I implemented what we talked about (a queue in `capture`! and fully removed `this._websites`. Please take a look and let me know what you think.

js/capture.js
```diff
@@ -9,16 +9,59 @@ const capture = {
...
+            changeInfo: changeInfo,
+            tab: tab
+          }
+        };
+        this.queue.push(eventDetails);
+        this.processNextEvent();
       });
   },
 
+  async processNextEvent(ignore = false) {
```
jonathanKingston: Let's add a comment why we are doing this here as it's not going to be obvious.

jonathanKingston: Like not just why we do ignore but the queue in general

js/capture.js
```diff
@@ -9,16 +9,59 @@ const capture = {
...
+          await this.sendFirstParty(
+            nextEvent.data.tabId,
+            nextEvent.data.changeInfo,
+            nextEvent.data.tab
+          );
+          break;
+        case 'sendThirdParty':
+          await this.sendThirdParty(nextEvent.data);
+          break;
+        default:
```
jonathanKingston: We should probably throw here rather than silently ignore.

jonathanKingston:
This is looking good, add some comments and double check it by testing and I think we are good to go. (I haven't played around with the updated copy though so will do that when you want me to merge it)

jonathanKingston:
Comments still needed and a throw.

Also the following code:
```
if (this.queue.length > 1) {
```
Should be greater than or equals to.
This currently means I can't load multiple first parties.

js/capture.js
biancadanforth: Okay I added some comments. Good point!

js/capture.js
biancadanforth: Okay I added an error to the default case.

biancadanforth:
Ok @jonathanKingston had to add some unglamorous fixes for some old but newly salient bugs in `capture.setThirdParty`. Let me know what you think. I'll let you decide whether the DNM flag should be removed.

jonathanKingston:
1. Can you instead check if document url is correct in this case? https://www.youtube.com/sw.js could be loading lots of it's own requests.
2. this is odd, going to check.

jonathanKingston:
- When we have tabId=-1 and no documentUrl, I don't think we can do anything here. (this seems to be super rare and might be related to prerendering or something)

Example of loading a sw:
```
"{
  "requestId": "1610",
  "url": "https://twitter.com/push_service_worker.js",
  "originUrl": "https://twitter.com/",
  "documentUrl": "https://twitter.com/",
  "method": "GET",
  "tabId": -1,
  "type": "script",
  "timeStamp": 1502443708292,
  "frameId": -1,
  "parentFrameId": -1,
  "fromCache": false,
  "ip": "104.244.42.1",
  "statusCode": 200,
  "statusLine": "HTTP/2.0 200 OK"
}"
```

Example of something a service worker loads:
```
"{
  "requestId": "1599",
  "url": "https://abs.twimg.com/favicons/favicon.ico",
  "originUrl": "https://twitter.com/mislav",
  "documentUrl": "https://twitter.com/mislav",
  "method": "GET",
  "tabId": -1,
  "type": "image",
  "timeStamp": 1502443707122,
  "frameId": -1,
  "parentFrameId": -1,
  "fromCache": true,
  "ip": "104.244.46.103",
  "statusCode": 200,
  "statusLine": "HTTP/2.0 200 OK"
}"
```

So we want to make sure `twimg.com` is a third party of `twitter.com` in this case. (I think however this is on the allowlist but you get the point).

js/capture.js
```diff
@@ -3,51 +3,109 @@
...
+      this.processingQueue = false;
+    }
+  },
+
   // Returns true if the request should be stored, otherwise false.
   shouldStore(tab) {
     const documentUrl = new URL(tab.url);
-    // ignore about:*, moz-extension:* & non-visible tabs (like dev tools)
+    // ignore about:* and moz-extension:*
     // also ignore third parties owned by first parties
```
jonathanKingston: `also ignore third parties owned by first parties`
Are we even checking for this any more?

jonathanKingston: I would rather do more work in `shouldStore` than make it not encompass the whole `shouldStore` algo, this would make it really hard for a newcomer to understand.

biancadanforth:
> Can you instead check if document url is correct in this case? https://www.youtube.com/sw.js could be loading lots of it's own requests.

@jonathanKingston , I'm not sure what you mean here... `response.documentUrl` is https://youtube.com/ in every case where `tabId` is -1 when I visit www.youtube.com. Could you clarify? Also what does a `tabId` of -1 mean in these cases generally?

js/capture.js
biancadanforth: > Are we even checking for this any more?

I meant to remove the comment at line 78; removed!

> I would rather do more work in shouldStore than make it not encompass the whole shouldStore algo, this would make it really hard for a newcomer to understand.

`shouldStore` requires a `tab` object, which we can't get if the `tabId` is -1 (that was what was throwing an error [above](https://github.com/electrolyfish/lightbeam-we/pull/115#issuecomment-321410437)). How would we add non-visible tabs back in?

js/capture.js
```diff
@@ -3,51 +3,109 @@
...
-        this.sendFirstParty(tabId, changeInfo, tab);
+        const eventDetails = {
+          type: 'sendFirstParty',
+          data: {
+            tabId: tabId,
+            changeInfo: changeInfo,
+            tab: tab
+          }
+        };
+        this.queue.push(eventDetails);
```
princiya: @biancadanforth you could change this to -->
```
data: { tabId, changeInfo, tab }
```

biancadanforth:
@jonathanKingston Can you give me some more specifics on how I could implement the Service Worker for testing capture?

Quoting you from slack:
> you should probably be able to load an image in the main tab or an iframe and it will go through the worker if you are capturing network requests. You can also make fetch requests too if the site supports CORS

So you're saying set up a worker like Archibald does in the [tutorial](https://github.com/jakearchibald/simple-serviceworker-tutorial) and load a third party image that it caches?

js/capture.js
biancadanforth: Good idea. :)

biancadanforth:
@jonathanKingston I've updated the capture code to no longer ignore non-visible tabs such as those from Service Workers. I also performed some experiments as you recommended regarding what values the `response` object returned by `webRequest.onResponseStarted` has for less common requests. You can find that experiment [here](https://github.com/biancadanforth/web-dev-experiments/pulls); though I had some questions about it and some limitations; which I ask you about [here](https://github.com/mozilla/lightbeam-we/pull/115#issuecomment-323102630).

jonathanKingston:
3. This is invalid code, when we discussed chaining iframes you can't just artificially make the relationships like this to stuff origins loading into other origins. This would break the webs security model too.

More like example.html
```html
<iframe src="https://localhost:80/frame.html"></iframe>
```

Which would load frame1.html:
```html
<iframe src="https://localhost:81/frame1.html"></iframe>
```

Which would load frame2.html:
```html
<iframe src="https://localhost:82/frame2.html"></iframe>
```

jonathanKingston:
2. Anything with a -1 tab id is likely going to be suspicious however I don't think we have any other way to distinguish these.

jonathanKingston:
1. Yeah exactly that, we should be able to make a fake site that loads a worker which has a document which loads other things. Either load from within the worker or make the script load it, however you might want to detect for when the worker has loaded `navigator.serviceWorker.ready` before you load the images/scripts from a third party so you can guarantee the worker loads it. You can't always guarantee just by loading a service worker that it will be active and on first load it most likely isn't loaded.

If you wait for worker load you can make some dummy js createElement and inject into the body.

So as Jakes example shows you can use the fetch event to capture what a HTML document is doing and this then acts as a proxy for the website (allowing the site to mess with the request or cache, do offline etc)

jonathanKingston:
> The original reason why we filtered out requests with tabId -1 was to ignore devTools tabs

Maybe this has changed since in Firefox I can't remember how we replicated this. We also wanted to filter out pre rendering which I still am seeing sometimes. I think there are other edge cases however most of these seem to have no document url.

jonathanKingston:
One thing I wanted to ensure with this research is that we are fully understanding what the properties the browser will be sending to the extension through these requests.

So it would be good to document what properties we get for each case that we listed in the meeting and perhaps update the MDN documentation if it isn't clear (which last time I checked it wasn't 100% clear on what the properties represent).

So for example we should get from this a certainty of how we will store all of the cases, we actually need to account for the cases where third parties load more third parties at some point so we can outline to the users of lightbeam what would be blocked by tracking protection (the graph will be bigger than just third parties on the list)

jonathanKingston:
One thing I wanted to ensure with this research is that we are fully understanding what the properties the browser will be sending to the extension through these requests.

So it would be good to document what properties we get for each case that we listed in the meeting and perhaps update the MDN documentation if it isn't clear (which last time I checked it wasn't 100% clear on what the properties represent).

So for example we should get from this a certainty of how we will store all of the cases, we actually need to account for the cases where third parties load more third parties at some point so we can outline to the users of lightbeam what would be blocked by tracking protection (the graph will be bigger than just third parties on the list)

^ I thought I had sent this last night :(

jonathanKingston:
Still looking into it but it looks like the patch is totally broken at the moment.

jonathanKingston:
Unfortunately we need to account for Firefox 55+56+57 at the moment, documentUrl is always undefined in stable at the moment.

jonathanKingston:
I tested this on: https://offline-news-service-worker.herokuapp.com/ and it's not storing at all. Sometimes it showed as a third party but not first.

jonathanKingston:
I made the following changes to make it work in 55 however it still doesn't fix that page above. We could do that as a follow up however it certainly would be a blocker to launch.

```
diff --git a/js/capture.js b/js/capture.js
index 88cd804..4e85a68 100644
--- a/js/capture.js
+++ b/js/capture.js
@@ -83,10 +87,11 @@ const capture = {
       documentUrl = new URL(tab.url);
       privateBrowsing = tab.incognito;
     } else {
       // browser.tabs.get throws an error for nonvisible tabs (tabId = -1)
       // but some non-visible tabs can make third party requests,
       // ex: Service Workers
-      documentUrl = new URL(info.documentUrl);
+      documentUrl = new URL(info.originUrl);
       privateBrowsing = false;
     }
       documentUrl = new URL(tab.url);
       privateBrowsing = tab.incognito;
     } else {
       // browser.tabs.get throws an error for nonvisible tabs (tabId = -1)
       // but some non-visible tabs can make third party requests,
       // ex: Service Workers
-      documentUrl = new URL(info.documentUrl);
+      documentUrl = new URL(info.originUrl);
       privateBrowsing = false;
     }
     // ignore about:*, moz-extension:*
@@ -101,8 +106,9 @@ const capture = {
 
   // capture third party requests
   async sendThirdParty(response) {
-    if (!response.documentUrl) {
-      // documentUrl is undefined for the first request from the browser to the
+    if (!response.originUrl) {
+      // originUrl is undefined for the first request from the browser to the
       // first party site
       return;
     }
@@ -110,20 +116,23 @@ const capture = {
     // @todo figure out why Web Extensions sometimes gives
     // undefined for response.originUrl
     const originUrl = response.originUrl ? new URL(response.originUrl) : '';
-    const documentUrl = new URL(response.documentUrl);
+    let documentHostname;
+    if (response.documentUrl) {
+      documentHostname = new URL(response.documentUrl).hostname;
+    }
     const targetUrl = new URL(response.url);
 
-    if (targetUrl.hostname !== documentUrl.hostname
+    if (targetUrl.hostname !== originUrl.hostname
       && await this.shouldStore(response)) {
       const data = {
-        document: documentUrl.hostname,
+        document: documentHostname,
         target: targetUrl.hostname,
         origin: originUrl.hostname,
         requestTime: response.timeStamp,
         firstParty: false
       };
       await store.setThirdParty(
-        documentUrl.hostname,
+        originUrl.hostname,
         targetUrl.hostname,
         data
       );
```

biancadanforth:
> Unfortunately we need to account for Firefox 55+56+57 at the moment, documentUrl is always undefined in stable at the moment.

I've been testing in Nightly (57) -- @jonathanKingston why do we need to account for pre-57 versions? Will look more closely at your comments soon.

biancadanforth:
> I tested this on: https://offline-news-service-worker.herokuapp.com/ and it's not storing at all. Sometimes it showed as a third party but not first.

@jonathanKingston Turns out I had at some point accidentally flipped the logic on `store.isNewWebsite`. I have fixed this in the latest commit. This may also have contributed to Issue #150, since only `store.setFirstParty` actually calls `isNewWebsite`. This would mean we'd be drawing a bunch of third party nodes to the page without drawing the first parties.

Aside/interesting observation: this site has 0 or 1 third party requests, depending on whether or not I manually unregister the SW in `about:debugging` before reloading the page.

Hopefully this next review of yours will pass!

jonathanKingston:
> Aside: The fact that what's returned as the value for documentUrl is different for different versions of Firefox makes it hard to know exactly how it is defined as you were hoping to find out.

Looking at your comments and initial testing I can't actually see the advantage we get from the document url. Let me know if that is an issue.

jonathanKingston:
Merging as I can't see any obvious issues now we still need an overarching review anyway and I raised #150 and #151 to do as a follow ups.

PR #118, opened 2017-08-07
Draw third parties as triangles

biancadanforth:
Updated the visualization to draw third parties as triangles. Also changed the colors to white. Both of these changes are to match the legacy UI.

I wanted to do this mostly for the sake of my intern presentation so that I could use screenshots from our version to compare and match with those of the legacy UI. If we're not ready to make this change (or don't want it), fine with me. :)

### DIFF
--- a/js/viz.js
+++ b/js/viz.js
@@ -105,12 +105,18 @@ const viz = {
     for (const d of this.nodes) {
       this.context.beginPath();
       this.context.moveTo(d.x, d.y);
-      this.context.arc(d.x, d.y, this.circleRadius, 0, 2 * Math.PI);
       if (d.firstParty) {
-        this.context.fillStyle = 'red';
+        this.context.arc(d.x, d.y, this.circleRadius, 0, 2 * Math.PI);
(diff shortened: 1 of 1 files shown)


js/viz.js
```diff
@@ -105,12 +105,18 @@ const viz = {
     for (const d of this.nodes) {
       this.context.beginPath();
       this.context.moveTo(d.x, d.y);
-      this.context.arc(d.x, d.y, this.circleRadius, 0, 2 * Math.PI);
       if (d.firstParty) {
-        this.context.fillStyle = 'red';
+        this.context.arc(d.x, d.y, this.circleRadius, 0, 2 * Math.PI);
       } else {
-        this.context.fillStyle = 'blue';
+        // draw third party as equilateral triangle
```
princiya: @biancadanforth please put this into a new function :) `drawFirstParty(x,y)` and `drawThirdParty(x,y)`

jonathanKingston:
Fixed @princiya feedback in #121

PR #133, opened 2017-08-09
added download button

pauljt:
I created the MOST important feature. Its perfect, you should definitely merge this.

### DIFF
--- a/js/lightbeam.js
+++ b/js/lightbeam.js
@@ -114,20 +114,24 @@ async function initLightBeam() {
   renderGraph(websites);
 
   const saveData = document.getElementById('save-data-button');
-  saveData.addEventListener('click', () => {
-    storeChild.getAll().then( data => {
-      const blob = new Blob([JSON.stringify(data,' ',2)],
-        {type : 'application/json'});
-      const url = window.URL.createObjectURL(blob);
-      browser.tabs.create({url: url});
-    });
+  saveData.addEventListener('click', async function () {
+    const data = await storeChild.getAll();
+    const blob = new Blob([JSON.stringify(data ,' ' , 2)],
+      {type : 'application/json'});
+    const url = window.URL.createObjectURL(blob);
+    const a = document.createElement('a');
+    a.href = url;
+    a.download = 'lightbeamData.json';
+    a.target = '_blank';
+    document.body.appendChild(a);
+    a.click();
+    // browser.tabs.create({url: url});
   });
 
   const resetData = document.getElementById('reset-data-button');
-  resetData.addEventListener('click', () => {
-    storeChild.reset().then(() => {
-      window.location.reload();
-    });
+  resetData.addEventListener('click', async function () {
+    await storeChild.reset();
+    window.location.reload();
   });
 }
 
--- a/js/lightbeam.js
+++ b/js/lightbeam.js
@@ -113,6 +113,16 @@ async function initLightBeam() {
   websites = await storeChild.getAll();
   renderGraph(websites);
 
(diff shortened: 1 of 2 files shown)


js/lightbeam.js
```diff
@@ -113,6 +113,16 @@ async function initLightBeam() {
   websites = await storeChild.getAll();
   renderGraph(websites);
 
+  const saveData = document.getElementById('save-data-button');
+  saveData.addEventListener('click', () => {
+    storeChild.getAll().then( data => {
```
jonathanKingston: Can we use await here instead of then?

Either use a named function callback to the event listener or change it to `'click', async function () {`

Can we look for the ESLint rule that requires consistent spacing too? So far we have functions with arguments that have spaces between but not spaces before `.then(data => {` rather than `.then( data =>`

js/lightbeam.js
```diff
@@ -113,6 +113,16 @@ async function initLightBeam() {
...
+      const blob = new Blob([JSON.stringify(data,' ',2)],
```
jonathanKingston: Nit: spaces here between arguments. Again not sure if there is a rule we can turn on.

js/lightbeam.js
```diff
@@ -113,6 +113,16 @@ async function initLightBeam() {
...
+        {type : 'application/json'});
+      const url = window.URL.createObjectURL(blob);
+      browser.tabs.create({url: url});
```
jonathanKingston: Can this download instead of opening a tab?

js/lightbeam.js
jonathanKingston: Also whilst you clean this up, you can clean the resetData then up which I did haha.

jonathanKingston:
Maybe move back to the new tab solution for the time being, we can file a follow up for making it download?

js/lightbeam.js
```diff
@@ -113,11 +113,25 @@ async function initLightBeam() {
...
+    const data = await storeChild.getAll();
+    const blob = new Blob([JSON.stringify(data ,' ' , 2)],
+      {type : 'application/json'});
+    const url = window.URL.createObjectURL(blob);
+    const a = document.createElement('a');
+    a.href = url;
+    a.download = 'lightbeamData.json';
+    a.target = '_blank';
+    document.body.appendChild(a);
+    a.click();
```
biancadanforth: @pauljt @jonathanKingston Why are we downloading by appending a link to the DOM and clicking it? Is this a common way (I honestly have never done this before)? Seems like the Web Extension downloads API would be the way to go.

Also when I try out your branch, Paul, the screen goes blank with a loading gif on the tab. This is the case when I remove `a.target='_blank'` as well.

jonathanKingston:
Hey @pauljt is this ready to review again? We can do some of the others as follow ups.

princiya:
@pauljt @jonathanKingston I wrote the following snippet for downloading and this works.

```
function downloadData() {
  const saveData = document.getElementById('save-data-button');
  saveData.addEventListener('click', async () => {
    const data = await storeChild.getAll();
    const blob = new Blob([JSON.stringify(data ,' ' , 2)],
      {type : 'application/json'});
    const url = window.URL.createObjectURL(blob);
    const downloading = browser.downloads.download({
      url : url,
      filename : 'lightbeamData.json',
      conflictAction : 'uniquify'
    });
    await downloading;
  });
}
```

Should I create a new PR?

princiya:
I created this PR #144 :)

Thanks @biancadanforth for the downloads api pointer!

Issue #126, opened 2017-08-08
Draw favicons on the graph


princiya:
If we have favicons, draw them on the graph for first party & third parties.

Issue #122, opened 2017-08-08
CSS styles for the selected node


princiya:
When a node is selected/dragged, the corresponding node needs to be highlighted.

![screen shot 2017-08-08 at 16 48 05](https://user-images.githubusercontent.com/8022693/29077922-6a71b4e2-7c59-11e7-8af1-23091c6355fd.png)

Issue #114, opened 2017-07-31
Fix tooltip direction


princiya:
![Tooltip](http://g.recordit.co/bd5Cl374IB.gif)

PR #112 takes care of responsiveness. But we need to fix the tooltip direction and avoid the horizontal scrollbar.

princiya:
Also, add styles to the tooltip with the arrow to match the current version.

PR #134, opened 2017-08-10
Updates

princiya:
In this PR:

Shadow

Updates

- [x]  Favicons #126 
- [x] Shadow highlight the dragged node #122 
- [x] Tooltip styles #114 
- [x] Tooltip position improvements #114

### DIFF
--- a/js/viz.js
+++ b/js/viz.js
@@ -9,9 +9,11 @@ const viz = {
   init(nodes, links) {
     const { width, height } = this.getDimensions('visualization');
     const { canvas, context } = this.createCanvas();
+    const { right: canvasRight } = canvas.getBoundingClientRect();
 
     this.canvas = canvas;
     this.context = context;
+    this.canvasRight = canvasRight;
     this.tooltip = document.getElementById('tooltip');
     this.circleRadius = this.circleRadius * this.scalingFactor;
     this.scale = (window.devicePixelRatio || 1) * this.scalingFactor;
@@ -183,19 +185,19 @@ const viz = {
   },
 
   getTooltipPosition(x, y) {
-    const { right: canvasRight } = this.canvas.getBoundingClientRect();
     const {
       height: tooltipHeight,
-      width: tooltipWidth,
-      right: tooltipRight
+      width: tooltipWidth
     } = this.tooltip.getBoundingClientRect();
     const top = y - tooltipHeight - this.circleRadius - 20;
+
     let left;
-    if ((tooltipRight + tooltipWidth) >= canvasRight) {
+    if ((x + tooltipWidth) >= this.canvasRight) {
       left = x - tooltipWidth - 20;
     } else {
       left = x - (tooltipWidth/2);
     }
+
     return {
       left,
       top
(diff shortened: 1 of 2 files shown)


js/viz.js
```diff
@@ -110,35 +110,105 @@ const viz = {
...
     this.context.lineTo(node.x + deltaX, node.y + deltaY);
   },
 
+  getTooltipPosition(x, y) {
+    const { right: canvasRight } = this.canvas.getBoundingClientRect();
+    const {
+      height: tooltipHeight,
+      width: tooltipWidth
+    } = this.tooltip.getBoundingClientRect();
+    const top = y - tooltipHeight - this.circleRadius - 20;
```
princiya: 20 here is for the tooltip arrow gutter (can be changed/omitted)

js/viz.js
```diff
@@ -110,35 +110,105 @@ const viz = {
...
+
+    let left;
+    if ((x + tooltipWidth) >= canvasRight
```
princiya: @jonathanKingston @biancadanforth please verify this math :)

js/viz.js
```diff
@@ -240,6 +311,10 @@ const viz = {
     return node;
   },
 
+  dragStart() {
+    d3.event.subject.shadow = true;
```
princiya: The current node highlight behaviour is as follows:
- Drag a node
- This is the active node and corresponding node has a shadow
- At drag end, this node isn't active anymore, and the shadow goes away

This behaviour is easy, as I know the active node via `d3.event.subject`.

@jonathanKingston @biancadanforth Do we need to highlight the node even after the drag end? (This is the behaviour in the old version)

In this case, we will have to do the following:
```
for(node in this.nodes) {
 node.shadow = false;
 if (node.hostname === d3.event.subject.hostname) {
   node.shadow = true;
 }
}
```
I don't want to iterate over all the nodes, this can be expensive? @jonathanKingston inputs welcome

js/viz.js
jonathanKingston: Can you make this a variable that is clear what it is for? (no comment needed just a clear var name)

princiya: sure! will do :)

js/viz.js
jonathanKingston: > Do we need to highlight the node even after the drag end? (This is the behaviour in the old version)

For now, no because we don't have the info section. This might change though.
Maybe file a follow up so we don't lose this comment though.

princiya: Sounds good :) Thanks

js/viz.js
jonathanKingston:
```
(x + tooltipWidth) >= canvasRight
|| (x + (tooltipWidth/2)) >= canvasRight
```
Surely the `/2` check will be redundant?

jonathanKingston: One of them is redundant unless you are running this code multiple times perhaps?

When `x + (tooltipWidth/2)) >= canvasRight` is true `(x + tooltipWidth) >= canvasRight` will always be true.

princiya: sometimes it just takes a fresh mind to realise that was indeed redundant! :D

Issue #129, opened 2017-08-08
Consider what UI features to add/remove


biancadanforth:
There are several features of the legacy Lightbeam that we have not implemented, for example:
* Site Info
* List View
* Tracking Protection
* Toggle Controls
* Save Data
... and others.

For any UI features not implemented, we should omit them from the UI (as opposed to disabling them) and create a separate issue in GitHub of the form: Consider adding X from legacy Lightbeam.

![screen shot 2017-08-08 at 9 17 42 am](https://user-images.githubusercontent.com/17437436/29082283-e84e4f82-7c19-11e7-92f6-d86e2c654266.png)

princiya:
@biancadanforth can we close this issue because of #139?

biancadanforth:
@princiya Yeah I think that's a good call. We have separate issues for implementing the other main features already. I can close this one myself!

PR #139, opened 2017-08-15
Hide unimplemented UI elements.

biancadanforth:
Added the `.unimplemented` class to unimplemented UI elements.

This PR is related to Issue #129 , where we discuss removing unimplemented features until they are ready.

I also removed the 'Uninstall Lightbeam' link.

### DIFF
--- a/css/style.css
+++ b/css/style.css
@@ -457,3 +457,11 @@ footer {
     grid-column-start: 1;
   }
 }
+
+.unimplemented {
+  visibility: hidden;
+}
+
+.unimplemented.list {
+  display: none;
+}
--- a/index.html
+++ b/index.html
@@ -25,7 +25,7 @@
       <div class="toggle-menu">
         <!-- @todo add click to toggle view between graph and list, highlighting active view -->
         <button class="active graph">Graph</button>
-        <button class="list">List</button>
+        <button class="list unimplemented">List</button>
       </div>
       <h2>Data</h2>
       <div>
@@ -36,9 +36,6 @@
       <div class="nav-links">
         <!-- @todo confirm feedback URL -->
         <a class="thumbs-up" href="mailto:lightbeam-feedback@mozilla.org">Give Us Feedback</a>
-        <br>
-        <!-- @todo confirm uninstall URL -->
-        <a href="https://support.mozilla.org/en-US/kb/disable-or-remove-add-ons">Uninstall Lightbeam</a>
       </div>
     </nav>
     </aside>
@@ -68,7 +65,7 @@
           </div>
         </dl>
         <!-- @todo add tracking protection and toggle functionality -->
-        <div class="tracking-protection">
+        <div class="tracking-protection unimplemented">
           <label for="tracking-protection-control">Tracking Protection</label>
           <div class="toggle-switch">
           <!-- @todo couple checkbox status (checked/unchecked) with toggle status -->
@@ -89,7 +86,7 @@
           <!-- @todo display list or graph view, update view based on filter and controls -->
           <!-- @todo remove the hard-coded width, height values -->
         </div>
-        <div class="vis-controls info-panel-controls">
+        <div class="vis-controls info-panel-controls unimplemented">
           <!-- @todo check purpose of website icon -->
           <button class="info-panel website"></button>
           <!-- @todo add help sidebar -->
@@ -102,7 +99,7 @@
         </div>
       </section>
       <footer>
-        <div class="footer-toggle">
+        <div class="footer-toggle unimplemented">
           <div class="footer-heading">
             <h2>Toggle Controls</h2>
           </div>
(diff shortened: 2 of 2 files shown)


princiya:
@biancadanforth My main concern here is agreeing on `display: flex` vs `display: grid`.

css/style.css
```diff
@@ -203,8 +206,8 @@ aside h2 {
 /* ----------- UI - Top Bar ----------- */
 
 main {
-  display: grid;
-  grid-template-rows: 10% 60% 30%;
```
princiya: I would say let's have `display: grid` here. Those toggles and filters will surely come soon after the internship.

css/style.css
```diff
@@ -266,12 +269,18 @@ input[type="checkbox"] {
 /* ----------- UI - Graph ----------- */
 
 .vis {
-  display: grid;
```
princiya: To me, this was a nice example of a multi-column layout and I feel `grid` is apt here too.

index.html
```diff
@@ -81,27 +81,29 @@
...
-        <div class="vis-controls info-panel-controls">
-          <!-- @todo check purpose of website icon -->
-          <button class="info-panel website"></button>
-          <!-- @todo add help sidebar -->
-          <button class="info-panel help"></button>
-          <!-- @todo add about sidebar -->
-          <button class="info-panel about"></button>
-        </div>
-        <div class="vis-content" id="visualization">
+        <div class="vis-header-wrapper">
```
princiya: The important CSS lesson I have learned so far, not to add <elements> just to match your layout! :) @jonathanKingston correct me here

css/style.css
```diff
@@ -455,5 +457,14 @@ footer {
...
+    padding-top: 20px;
   }
+
+  .footer-toggle-buttons {
+    grid-template-columns: 1fr;
+  }
+}
+
+.unimplemented {
+  display: none;
```
princiya: how about changing this to `visibility: hidden` instead? That way, the DOM structure isn't broken and probably we won't have to restructure the UI, unless there are strong reasons to change from grid to flex :)

index.html
jonathanKingston: Agreed do we need `display: contents;` or something? Sometimes it is unavoidable though (haven't actually looked). I think only ff works with display contents.

css/style.css
jonathanKingston: This seems a good idea for some cases at least. I guess there might be exceptions.

jonathanKingston:
> What URL should we use for 'Give Us Feedback'? 

We should file a follow up for this to use Discourse as @groovecoder thinks that is the best direction.

css/style.css
biancadanforth: This is actually one of the few areas where I think flexbox is more appropriate for sure.

The reason why I changed this to `display: flex` is because otherwise there was uneven whitespace underneath the toggles and filters relative to the rest of the main content area. Giving a height for the footer of 30% instead of using flexbox creates this unevenness. Even using fractional units (ex: 1fr with grid) still says: Make this element X relative to the size of other elements/the height of the page. What we really want is to say: place it as far down the page as possible given its size. `display: flex`, `flex-direction: column` and `justify-content: space-between` allows us to do that. Doing this allows us not only to have even whitespace around the main content area, but also that extra space the footer was taking up in grid is now used to make the visualization area larger.

Before:
screen shot 2017-08-16 at 8 28 29 am

After (with `.unimplemented` class removed on footer)
screen shot 2017-08-16 at 8 31 07 am

css/style.css
biancadanforth: This also suffers from the same issue as I mentioned above. With Grid, all we can say is: take up X amount of height relative to other rows in the grid or the height of the page. What we really want is this `.vis` element's child elements (rows in this case) to only take up as much space as they need to given their content and padding, and we want the `#visualization` area to absorb the rest of the space.

Inspecting this grid in Nightly (with its cool Layout inspector!) shows us that the headers at the top are taking up more space than they need to be, taking away space from the visualization area.

screen shot 2017-08-16 at 8 38 26 am

princiya: current `.vis` has `1fr 2fr`. We can change this ratio anytime, `1fr 3fr` gives more space to `#visualization`.

index.html
```diff
@@ -25,20 +25,17 @@
...
         <!-- @todo add click to toggle view between graph and list, highlighting active view -->
         <button class="active graph">Graph</button>
-        <button class="list">List</button>
+        <button class="list unimplemented">List</button>
       </div>
       <h2>Data</h2>
       <div>
         <!-- @todo add click to save data -->
-        <button class="download">Save Data</button>
+        <button class="download unimplemented">Save Data</button>
```
princiya: We have this feature in PR #144. This is implemented ;)

PR #155, opened 2017-08-21
Fix legacy Firefox versions and missing graph nodes not connected to …

### LINKED ISSUE #150: Disconnected third parties on the graph
In @biancadanforth's patch I am still seeing nodes that are triangle and disconnected from any other node, this should never happen.

### LINKED ISSUE #151: Fix older version of capture
See comments here:
https://github.com/mozilla/lightbeam-we/pull/115#issuecomment-323539679

jonathanKingston:
…their first parties. Fixes #150 Fixes #151

### DIFF
--- a/js/capture.js
+++ b/js/capture.js
@@ -76,9 +76,8 @@ const capture = {
   // response (from setThirdParty) object
   async shouldStore(info) {
     const tabId = info.id || info.tabId;
-    const visibleTab = tabId !== browser.tabs.TAB_ID_NONE;
     let tab, documentUrl, privateBrowsing;
-    if (visibleTab) {
+    if (this.isVisibleTab(tabId)) {
       tab = await browser.tabs.get(tabId);
       documentUrl = new URL(tab.url);
       privateBrowsing = tab.incognito;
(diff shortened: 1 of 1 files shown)


jonathanKingston:
//cc heads up @princiya and @biancadanforth I'm going to merge this as it fixes the issues I found. However in our storage checks we need to make sure this is working correctly.

princiya:
what was the cause for disconnected nodes?

jonathanKingston:
From slack:

documentUrl is just a frame that loaded the request
tab url is always the top level frame
except when we don't have a tab, which it seems origin url is most accurate

biancadanforth:
@jonathanKingston on slack:
> Tab.url always captures the top level frame, whereas documentURL may not be the parent frame all the time

One example is when you load another page in an iframe, then for the requests to load that page, documentUrl points to the iframe page.

I still don't see why `originUrl` is better than `documentUrl` or why we were seeing disconnected nodes?

biancadanforth:
From @jonathanKingston in slack:

> I still don't see why originUrl is better than documentUrl

origin is more right for when we don't have a tab
and works in 55
origin points to it's loading origin, which could be a script that loaded an iframe
so document would be the iframe, origin would be the js

> "And why were we seeing disconnected nodes from the capture code?"

From initial testing it looked like we had the third parties not loaded from a document so a script loaded a third party so it wasn't being associated correclty

Issue #147, opened 2017-08-19
Flickering of favicons on the graph when thy are moving


jonathanKingston:
STR
- Load some pages like google.co.uk urban dictionary etc.
- Drag the nodes on the graph

Expected
No flickering

Actual
Images apprear to be redrawn for every drag event, sometimes I also see multiple of the same favicon too

jonathanKingston:
This is now much worse based on the code pushed in #158 the flickering and rendering multiple nodes feel the same issue however feel free to split if it isn't. The ghosting/multiple of the same nodes is the p1 component here.

jonathanKingston:
I'm going to close this, we do however have a follow up of #212

PR #158, opened 2017-08-21
Improve graph forces

princiya:
Screenshot

Here is the summary of this PR:

```
simulateForce() {
    let simulation;

    if (!this.simulation) {
      simulation = d3.forceSimulation(this.nodes);
      this.registerSimulationForces(simulation);
    } else {
      simulation = this.simulation;
      simulation.nodes(this.nodes);
    }
    this.registerLinkForce(simulation);
    this.manualTick(simulation);
    return simulation;
  }
```
- On page load, simulation is initialised and the necessary forces are registered.
- For every draw (i.e. when new nodes/links are added), the `linkForce` is re-initialised.
- Because of the dynamic nature of the graph, manual tick is used.
   - When there are too many nodes, the graph constantly tries to to map to the registered forces (in our case, for collisions and to centre the graph) and sometimes ends up in an infinite (long) animated loop.
  - Also, when there are lot of nodes, it takes a while for the simulation to stabilise and favicons aren't loaded until the simulation has reached equilibrium.
- In terms of the whole graph updating when new nodes are added, it is to make sure there are no collisions and the graph is always centred.

---

```
manualTick(simulation) {
    simulation.alphaTarget(0.1);
    for (let i = 0; i < this.tickCount; i++) {
      simulation.tick();
    }
    simulation.alphaTarget(0);
    simulation.stop();
  }
```

The soft animation seen on the graph is because of the simulation and it can be controlled with the value of `alphaTarget`. At each tick, it is ensured to meet the conditions of the above registered forces. This is why, the graph constantly rotates/repels/attracts/non-collides based on the forces registered.

---

```
registerSimulationForces(simulation) {
    const centerForce = d3.forceCenter(this.width/2, this.height/2);
    simulation.force('center', centerForce);

    const forceX = d3.forceX(this.width/2);
    simulation.force('x', forceX);

    const forceY = d3.forceY(this.height/2);
    simulation.force('y', forceY);

    const chargeForce = d3.forceManyBody();
    simulation.force('charge', chargeForce);

    const collisionForce = d3.forceCollide(this.collisionRadius);
    simulation.force('collide', collisionForce);
  }
```
This is one time initialisation, during the page load and the simulation ensures these conditions are met for every tick. The graph alters only when it detects collision or to centre the graph etc. based on the above values set.

`centerForce` is used to centre the graph w.r.to the display area.

`forceX` and `forceY` ensure newly added nodes are w.r.to the centre of the graph.
This is an enhancement in this PR, earlier the newly added nodes started at `<0,0>`.

`collisionForce` ensures there is no collision between nodes. Try to drag a node towards another node, and because of this `collisionForce` the graph tries to redraw itself to ensure there is no collision.

---

### DIFF
--- a/js/viz.js
+++ b/js/viz.js
@@ -16,6 +16,7 @@ const viz = {
     this.context = context;
     this.tooltip = document.getElementById('tooltip');
     this.circleRadius = this.circleRadius * this.scalingFactor;
+    this.collisionRadius = this.collisionRadius * this.scalingFactor;
     this.scale = (window.devicePixelRatio || 1) * this.scalingFactor;
     this.transform = d3.zoomIdentity;
 
@@ -28,59 +29,60 @@ const viz = {
     this.nodes = nodes;
     this.links = links;
 
-    this.simulate();
+    this.simulateForce();
     this.drawOnCanvas();
   },
 
-  simulate() {
-    this.simulation = this.simulateForce();
-  },
-
   simulateForce() {
-    let simulation;
-
     if (!this.simulation) {
-      simulation = d3.forceSimulation(this.nodes);
-      this.registerSimulationForces(simulation);
+      this.simulation = d3.forceSimulation(this.nodes);
+      this.registerSimulationForces();
     } else {
-      simulation = this.simulation;
-      simulation.nodes(this.nodes);
+      this.simulation.nodes(this.nodes);
     }
-    this.registerLinkForce(simulation);
-    this.manualTick(simulation);
-    return simulation;
+    this.registerLinkForce();
+    this.manualTick();
   },
 
-  manualTick(simulation) {
-    simulation.alphaTarget(0.1);
+  manualTick() {
+    this.simulation.alphaTarget(0.1);
     for (let i = 0; i < this.tickCount; i++) {
-      simulation.tick();
+      this.simulation.tick();
     }
-    simulation.alphaTarget(0);
-    simulation.stop();
+    this.stopSimulation();
   },
 
-  registerLinkForce(simulation) {
+  restartSimulation() {
+    this.simulation.alphaTarget(0.1);
+    this.simulation.restart();
+  },
+
+  stopSimulation() {
+    this.simulation.alphaTarget(0);
+    this.simulation.stop();
+  },
+
+  registerLinkForce() {
     const linkForce = d3.forceLink(this.links);
     linkForce.id((d) => d.hostname);
-    simulation.force('link', linkForce);
+    this.simulation.force('link', linkForce);
   },
 
-  registerSimulationForces(simulation) {
+  registerSimulationForces() {
     const centerForce = d3.forceCenter(this.width/2, this.height/2);
-    simulation.force('center', centerForce);
+    this.simulation.force('center', centerForce);
 
     const forceX = d3.forceX(this.width/2);
-    simulation.force('x', forceX);
+    this.simulation.force('x', forceX);
 
     const forceY = d3.forceY(this.height/2);
-    simulation.force('y', forceY);
+    this.simulation.force('y', forceY);
 
     const chargeForce = d3.forceManyBody();
-    simulation.force('charge', chargeForce);
+    this.simulation.force('charge', chargeForce);
 
     const collisionForce = d3.forceCollide(this.collisionRadius);
-    simulation.force('collide', collisionForce);
+    this.simulation.force('collide', collisionForce);
   },
 
   createCanvas() {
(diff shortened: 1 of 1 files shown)


js/viz.js
```diff
@@ -47,16 +47,25 @@ const viz = {
...
     simulation.force('center', centerForce);
 
+    const forceX = d3.forceX(this.width/2);
+    forceX.strength(0.3);
+    simulation.force('x', forceX);
+
+    const forceY = d3.forceY(this.height/2);
+    forceY.strength(0.3);
+    simulation.force('y', forceY);
+
```
princiya: As seen from the above gifs, before this PR, the newly added nodes started from <0,0>. It is now improved to start from position `<x,y>`. This way nodes are always clustered toward the center.

js/viz.js
```diff
@@ -47,16 +47,25 @@ const viz = {
...
     centerForce.x(this.width/2);
     centerForce.y(this.height/2);
     simulation.force('center', centerForce);
 
+    const forceX = d3.forceX(this.width/2);
+    forceX.strength(0.3);
+    simulation.force('x', forceX);
+
+    const forceY = d3.forceY(this.height/2);
+    forceY.strength(0.3);
```
jonathanKingston: Can you make this a descriptive property? it's used twice.

js/viz.js
```diff
@@ -47,16 +47,25 @@ const viz = {
 
     const linkForce = d3.forceLink(this.links);
     linkForce.id((d) => d.hostname);
-    linkForce.distance(100);
+    linkForce.distance(50);
```
jonathanKingston: Can you make this a descriptive property also?

js/viz.js
```diff
@@ -45,22 +49,44 @@ const viz = {
       simulation.nodes(this.nodes);
     }
 
+    this.registerSimulationForces(simulation);
+
+    simulation.alphaTarget(1);
+    for (let i = 0; i < 10; i++) {
+      simulation.tick();
+    }
```
princiya: manually updating `tick` to avoid the node collisions. @jonathanKingston let me know what you think.

princiya:
PR is ready, @jonathanKingston let me know if the graph behaves as expected.

js/viz.js
jonathanKingston: This 10 constant seems pretty arbitrary. This should also be a descriptive constant somewhere.
However is there any way to keep ticking until there are no more changes?

js/viz.js
```diff
@@ -325,13 +351,11 @@ const viz = {
 
   dragEnd() {
     d3.event.subject.shadow = false;
-    this.endEvent();
```
jonathanKingston: I think this solves some of the issues however on update the whole graph is still recalculated, so perhaps this isn't the right change?

jonathanKingston: Also when I move a first party I was kind of expecting the third parties to move closer to where the first party went to also.

jonathanKingston:
Also can you rebase this patch, it should fix the node detached issue etc.

js/viz.js
princiya: @jonathanKingston > on update the whole graph is still recalculated

Do you mean, that when new nodes come in, the existing graph shouldn't change? And the graph should change/recalculated only during page refresh?

js/viz.js
princiya: we can register event handler for `tick` as below:
`simulation.on('tick', () => doSomething);`
This wasn't behaving as expected, the graph was starting from 0,0 and not taking into account the simulationForces etc. So I resolved to having manual tick as a temporary solution.

js/viz.js
jonathanKingston: I would like it not to throw the whole graph around yeah, things can move but recalculating the whole graph shouldn't happen within one paint certainly.

jonathanKingston: The current lightbeam stops moving after a while so doesn't keep repelling and rendering again.

js/viz.js
jonathanKingston: Do we need to initialise the graph positions on page load then?

biancadanforth:
@princiya Awesome to see the third parties moving about and influencing each other! Though seems like the first parties are more stubborn? They don't seem to move at all in relation.

firstpartiesdontmove

jonathanKingston:
This looks great as mentioned #147 is a follow up here. If @biancadanforth thinks her comment is an issue file an issue however to me it looks intended.

Thanks and well done!

biancadanforth:
Princiya and I talked about it in slack and she fixed it in the commit following the comment; looks awesome!

PR #160, opened 2017-08-21
Restructure lightbeam.js similar to other scripts

biancadanforth:
@jonathanKingston Good idea to make this a separate PR to the dynamic variables...

Basically I just made the `lightbeam.js` script of the same form as `capture.js`, `store.js`, etc., where `lightbeam` is a global object with methods. I have not modified any methods, just moved things around and renamed the anonymous function that passed an update callback to `storeChild` as `redraw`.

### DIFF
--- a/.eslintrc.json
+++ b/.eslintrc.json
@@ -9,6 +9,7 @@
         "store": true,
         "storeChild": true,
         "capture": true,
+        "lightbeam": true,
         "viz": true,
         "d3": true,
         "Dexie": true
--- a/js/lightbeam.js
+++ b/js/lightbeam.js
@@ -1,61 +1,31 @@
-// transforms the object of nested objects 'websites' into a
-// usable format for d3
+const lightbeam = {
+  websites: {},
 
-/*
-  websites is expected to match this format:
-  {
-    "www.firstpartydomain.com": {
-      favicon: "http://blah...",
-      firstParty: true,
-      firstPartyHostnames: false,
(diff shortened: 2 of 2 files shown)


jonathanKingston:
Can you outline the changes here this is a pretty noisy diff. Could you
remove the WIP and fixes as I think this should just be a cleanup bug
instead and not mixed with a much bigger patch? Refractors should be
treated in isolation and with as much rigour as anything else.
Feel free to raise a bug and mark the fixes as that bug id.

jonathanKingston:
This looks ok, can you change `lightbeam` in `init()` to use `this` to be
consistent.

jonathanKingston:
Which probably will require the onload to be an anon arrow function that
calls init().

biancadanforth:
Oh I forgot to mention 'this' binding points to window there after the
implicit reassignment on the previous line, so directly using 'this'
doesn't work... Any suggestions?

jonathanKingston:
Yeah my follow up email got lost perhaps:

window.onload = () => {
  Lightbeam.init();
};

biancadanforth:
Okay @jonathanKingston , I updated the `onload` event to use an arrow function in the callback. Ready for you!

jonathanKingston:
Useful tip @biancadanforth checking this branch out and doing `git show <commit-id> -w` makes the commit super readable. I dunno why github defaults to showing the whitespace changes certainly for JS files at least.

PR #164, opened 2017-08-22
Increase graph area

princiya:
Image

### DIFF
--- a/css/style.css
+++ b/css/style.css
@@ -278,6 +278,7 @@ input[type="checkbox"] {
   bottom: 0;
   left: 0;
   right: 0;
+  overflow: hidden;
 }
 
 #tooltip {
--- a/css/style.css
+++ b/css/style.css
@@ -254,12 +254,14 @@ input[type="checkbox"] {
 
 .vis {
   display: grid;
-  grid-template-rows: 1fr 2fr;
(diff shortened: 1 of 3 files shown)


css/style.css
```diff
@@ -271,7 +273,11 @@ input[type="checkbox"] {
 }
 
 .vis-content {
-  position: relative;
+  position: absolute;
+  top: 0;
+  bottom: 0;
+  left: 0;
+  right: 0;
```
jonathanKingston: Can you make this overflow: hidden;  also or solve why there are scroll bars perhaps?

js/viz.js
```diff
@@ -86,8 +86,8 @@ const viz = {
     this.context.scale(this.scale, this.scale);
   },
 
-  getDimensions(id) {
-    const element = document.getElementById(id);
+  getDimensions() {
+    const element = document.body;
     const { width, height } = element.getBoundingClientRect();
```
princiya: because we are calculating width, height w.r.to the body element, and the graph uses width/2, height/2 as the centre point, we have the following issue for resize:

screen shot 2017-08-22 at 14 32 33

The graph is not really centred now w.r.to the vis area. Will file a follow-up issue..

PR #169, opened 2017-08-22
Make vars dynamic. Fixes #65.

### LINKED ISSUE #65: Make vars dynamic

biancadanforth:
Added a few new methods to lightbeam.js and store.js to fetch these values from storage to initialize them, and then increment numVisitedSites and numThirdParties when appropriate on each redraw() in lightbeam.js.

It should be noted that for some reason [Dexie filtering does not support boolean values](http://dexie.org/docs/WhereClause/WhereClause.equals().html) for primary keys or indices, so I had to convert the `firstParty` and `isVisible` keys to use `0` for `false` and `1` for `true`...

### DIFF
--- a/index.html
+++ b/index.html
@@ -43,23 +43,20 @@
         <dl class="info-headings">
           <div>
             <dt>Data Gathered Since</dt>
-              <!-- @todo get date-start from store -->
-            <dd id="date-start">
-              <time datetime="2017-06-13T11:00">June 13, 2017</time>
+            <dd>
+              <time id="data-gathered-since" datetime="2017-06-13T11:00"></time>
             </dd>
           </div>
           <div>
             <dt>You Have Visited</dt>
-            <!-- @todo get num-sites-visited from store -->
             <dd>
-              <var id="num-sites-visited">6</var> Sites
+              <var id="num-sites-visited">0</var> Sites
             </dd>
           </div>
           <div>
             <dt>You Have Connected With</dt>
-            <!-- @todo get num-third-parties from store -->
             <dd>
-              <var id="num-third-parties">16</var> Third Party Sites
+              <var id="num-third-parties">0</var> Third Party Sites
             </dd>
           </div>
         </dl>
--- a/index.html
+++ b/index.html
@@ -44,19 +44,19 @@
           <div>
             <dt>Data Gathered Since</dt>
             <dd>
-              <time id="data-gathered-since" datetime="2017-06-13T11:00"></time>
+              <time id="data-gathered-since" datetime=""></time>
             </dd>
           </div>
           <div>
             <dt>You Have Visited</dt>
             <dd>
-              <var id="num-sites-visited">0</var> Sites
+              <var id="num-first-parties"></var>
             </dd>
           </div>
           <div>
             <dt>You Have Connected With</dt>
             <dd>
-              <var id="num-third-parties">0</var> Third Party Sites
+              <var id="num-third-parties"></var>
             </dd>
           </div>
         </dl>
--- a/js/capture.js
+++ b/js/capture.js
@@ -135,7 +135,7 @@ const capture = {
         target: targetUrl.hostname,
         origin: originUrl.hostname,
         requestTime: response.timeStamp,
-        firstParty: 0
+        firstParty: false
       };
       await store.setThirdParty(
         firstPartyUrl.hostname,
@@ -151,7 +151,8 @@ const capture = {
     if (tab.status === 'complete' && await this.shouldStore(tab)) {
       const data = {
         faviconUrl: tab.favIconUrl,
-        firstParty: 1
+        firstParty: true,
+        requestTime: Date.now()
       };
       await store.setFirstParty(documentUrl.hostname, data);
     }
--- a/js/capture.js
+++ b/js/capture.js
@@ -135,7 +135,7 @@ const capture = {
         target: targetUrl.hostname,
         origin: originUrl.hostname,
         requestTime: response.timeStamp,
-        firstParty: false
+        firstParty: 0
       };
       await store.setThirdParty(
         firstPartyUrl.hostname,
@@ -151,7 +151,7 @@ const capture = {
     if (tab.status === 'complete' && await this.shouldStore(tab)) {
       const data = {
         faviconUrl: tab.favIconUrl,
-        firstParty: true
+        firstParty: 1
       };
       await store.setFirstParty(documentUrl.hostname, data);
     }
--- a/js/lightbeam.js
+++ b/js/lightbeam.js
@@ -1,7 +1,7 @@
 const lightbeam = {
   websites: {},
   dataGatheredSince: '',
-  numSitesVisited: 0,
+  numFirstParties: 0,
   numThirdParties: 0,
 
   async init() {
@@ -25,47 +25,65 @@ const lightbeam = {
   },
 
   // Called from init() (isFirstParty = undefined)
-  // and redraw() (isFirstParty = 1 or 0).
+  // and redraw() (isFirstParty = true or false).
   async updateVars(isFirstParty) {
-    const numSitesVisited = document.getElementById('num-sites-visited');
+    const numFirstParties = document.getElementById('num-first-parties');
     const numThirdParties = document.getElementById('num-third-parties');
 
     // initialize dynamic vars from storage
     if (!this.dataGatheredSince) {
-      this.dataGatheredSince = await this.getDataGatheredSince();
+      const { dateStr, fullDateTime } = await this.getDataGatheredSince();
+      this.dataGatheredSince = dateStr;
       const dataGatheredSince = document.getElementById('data-gathered-since');
       dataGatheredSince.textContent = this.dataGatheredSince;
+      dataGatheredSince.setAttribute('datetime', fullDateTime);
     }
     if (isFirstParty === undefined) {
-      this.numSitesVisited = await this.getNumSitesVisited();
-      numSitesVisited.textContent = this.numSitesVisited;
+      this.numFirstParties = await this.getNumFirstParties();
+      numFirstParties.textContent
+        = (this.numFirstParties === 0) ? '' : `${this.numFirstParties} Sites`;
       this.numThirdParties = await this.getNumThirdParties();
-      numThirdParties.textContent = this.numThirdParties;
+      numThirdParties.textContent
+        = (this.numThirdParties === 0) ? ''
+          : `${this.numThirdParties} Third Party Sites`;
       return;
     }
 
     // update on redraw
     if (isFirstParty) {
-      this.numSitesVisited++;
-      numSitesVisited.textContent = this.numSitesVisited;
+      this.numFirstParties++;
+      numFirstParties.textContent
+        = (this.numFirstParties === 0) ? '' : `${this.numFirstParties} Sites`;
     } else {
       this.numThirdParties++;
-      numThirdParties.textContent = this.numThirdParties;
+      numThirdParties.textContent
+        = (this.numThirdParties === 0) ? ''
+          : `${this.numThirdParties} Third Party Sites`;
     }
   },
 
   async getDataGatheredSince() {
-    const oldestDateUnixTime = await storeChild.getOldestDate();
-    if (!oldestDateUnixTime) {
+    const firstRequestUnixTime = await storeChild.getFirstRequestTime();
+    if (!firstRequestUnixTime) {
       return '';
     }
     // reformat unix time
-    const fullDateTime = new Date(oldestDateUnixTime);
-    return fullDateTime.toDateString();
+    let fullDateTime = new Date(firstRequestUnixTime);
+    let dateStr = fullDateTime.toDateString();
+    // remove day of the week
+    const dateArr = dateStr.split(' ');
+    dateArr.shift();
+    dateStr = dateArr.join(' ');
+    // ISO string used for datetime attribute on <time>
+    fullDateTime = fullDateTime.toISOString();
+    return {
+      dateStr,
+      fullDateTime
+    };
   },
 
-  async getNumSitesVisited() {
-    return await storeChild.getNumSitesVisited();
+  async getNumFirstParties() {
+    return await storeChild.getNumFirstParties();
   },
 
   async getNumThirdParties() {
--- a/js/lightbeam.js
+++ b/js/lightbeam.js
@@ -1,6 +1,6 @@
 const lightbeam = {
   websites: {},
-  dataGatheredSince: '',
+  dataGatheredSince: null,
   numFirstParties: 0,
   numThirdParties: 0,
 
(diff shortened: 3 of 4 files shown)


index.html
```diff
@@ -44,23 +44,20 @@
         <dl class="info-headings">
           <div>
             <dt>Data Gathered Since</dt>
-              <!-- @todo get date-start from store -->
-            <dd id="date-start">
-              <time datetime="2017-06-13T11:00">June 13, 2017</time>
+            <dd>
+              <time id="data-gathered-since" datetime="2017-06-13T11:00"></time>
```
jonathanKingston: The datetime property here needs to be dynamic.

index.html
```diff
@@ -44,23 +44,20 @@
...
+            <dd>
+              <time id="data-gathered-since" datetime="2017-06-13T11:00"></time>
             </dd>
           </div>
           <div>
             <dt>You Have Visited</dt>
-            <!-- @todo get num-sites-visited from store -->
             <dd>
-              <var id="num-sites-visited">6</var> Sites
+              <var id="num-sites-visited">0</var> Sites
```
jonathanKingston: I think the default would be better blank, perhaps not even showing the word "sites" until it's populated. What do you think?

js/capture.js
```diff
@@ -135,7 +135,7 @@ const capture = {
         target: targetUrl.hostname,
         origin: originUrl.hostname,
         requestTime: response.timeStamp,
-        firstParty: false
+        firstParty: 0
```
jonathanKingston: I think this should remain boolean here and changed in store.js. Perhaps in store.js add a comment that indexes can't be on boolean when you make that conversion too?

js/store.js
```diff
@@ -360,6 +363,29 @@ const store = {
     // empty out request processing queue
     capture.queue = [];
     return await this.db.websites.clear();
+  },
+
+  async getOldestDate() {
+    const oldestSite = await this.db.websites.orderBy('requestTime').first();
+    if (!oldestSite) {
+      return '';
```
jonathanKingston: I think this should return false, this blank string is actually presentation. Looks like you already would cope with that in lightbeam anyway.

js/store.js
```diff
@@ -360,6 +363,29 @@ const store = {
     // empty out request processing queue
     capture.queue = [];
     return await this.db.websites.clear();
+  },
+
+  async getOldestDate() {
```
jonathanKingston: Perhaps something descriptive of what data is oldest would be better `getFirstRequestTime`?

js/store.js
```diff
@@ -360,6 +363,29 @@ const store = {
...
+    const oldestSite = await this.db.websites.orderBy('requestTime').first();
+    if (!oldestSite) {
+      return '';
+    }
+    return oldestSite.requestTime;
+  },
+
+  async getNumSitesVisited() {
+    // Dexie does not accept boolean values; using 0/1 instead
```
jonathanKingston: Change this comment to be indexeddb indexes can't be boolean. Perhaps put it in one place rather than copy it in many?

js/store.js
```diff
@@ -360,6 +363,29 @@ const store = {
...
+
+  async getOldestDate() {
+    const oldestSite = await this.db.websites.orderBy('requestTime').first();
+    if (!oldestSite) {
+      return '';
+    }
+    return oldestSite.requestTime;
+  },
+
+  async getNumSitesVisited() {
```
jonathanKingston: Maybe this should be `getNumFirstParties` and leave "sites visited" as presentation only language rather than in the code.

js/store.js
```diff
@@ -298,10 +301,10 @@ const store = {
...
           // hide third party
-          thirdParty['isVisible'] = false;
+          thirdParty['isVisible'] = 0;
         } else {
           // show third party; it either became visible or is brand new
-          thirdParty['isVisible'] = true;
+          thirdParty['isVisible'] = 1;
           isNewThirdParty = true;
           for (let i = 0; i < thirdParty['firstPartyHostnames'].length; i++) {
             const firstPartyHostname = thirdParty['firstPartyHostnames'][i];
```
jonathanKingston: The existing code does this:
```
for (const key in data) {
  thirdParty[key] = data[key];
}
```

We will need to do smarter things for merging this for keys we don't want to just update. So `requestTime` should likely be stored as: `firstRequestTime` and `latestRequestTime` (adding both now to account for clearing of data would be worthwhile)

js/lightbeam.js
```diff
@@ -20,6 +24,54 @@ const lightbeam = {
...
+    const numSitesVisited = document.getElementById('num-sites-visited');
+    const numThirdParties = document.getElementById('num-third-parties');
+
+    // initialize dynamic vars from storage
+    if (!this.dataGatheredSince) {
+      this.dataGatheredSince = await this.getDataGatheredSince();
+      const dataGatheredSince = document.getElementById('data-gathered-since');
+      dataGatheredSince.textContent = this.dataGatheredSince;
+    }
+    if (isFirstParty === undefined) {
```
jonathanKingston: I'm not a massive fan of this being undefined, however too tired to think of anything better right now.

index.html
biancadanforth: Okay; fixed this thank you. I converted the Date object to an ISO string that looks to be an acceptable format, `2017-08-23T17:21:54.495Z` as an example value for the `datetime` attribute on the `time` element...

index.html
biancadanforth: I agree; changed this. I had to modify the flexbox style on `.top-bar` to `align-items: flex-start` to keep content from shifting when the text is added.

js/capture.js
biancadanforth: Okay I kept all boolean data types as is and only change them in the `setWebsite` method in store with a single comment about it.

js/store.js
biancadanforth: Okay I have updated `setWebsite` to do this for both first and third parties.

js/store.js
biancadanforth: Sure! Done.

js/store.js
biancadanforth: Good point; redundant. Better to make it a boolean data type.

js/store.js
biancadanforth: Okay sounds good. Done.

js/store.js
biancadanforth: Done!

biancadanforth:
@jonathanKingston This is ready for a look-see. :) Have a couple questions in the most recent commit description:

Should I worry about the <dl> elements in `.top-bar` shifting when there is and isn't content in these fields? Was thinking I could do a pseudoelement and remove it once they're populated.

Also, if you know a better way in lightbeam.updateVars to distinguish between initialization and update other than `isFirstParty === undefined`, please let me know!

js/lightbeam.js
```diff
@@ -1,10 +1,14 @@
 const lightbeam = {
   websites: {},
+  dataGatheredSince: '',
```
jonathanKingston: I would make this null and at presentation time turn this into a blank string to be clear the difference between the data itself and the presentation.

js/lightbeam.js
```diff
@@ -20,6 +24,72 @@ const lightbeam = {
...
+      dataGatheredSince.textContent = this.dataGatheredSince;
+      dataGatheredSince.setAttribute('datetime', fullDateTime);
+    }
+    if (isFirstParty === undefined) {
+      this.numFirstParties = await this.getNumFirstParties();
+      numFirstParties.textContent
+        = (this.numFirstParties === 0) ? '' : `${this.numFirstParties} Sites`;
+      this.numThirdParties = await this.getNumThirdParties();
+      numThirdParties.textContent
+        = (this.numThirdParties === 0) ? ''
```
jonathanKingston: Multiline ternary ifs are pretty hard to read, can we use an if statement here? Otherwise save the string in a var.
```
const thing = `ssss`;
numThirdParties.textContent = (!this.numThirdParties) ? '' : thing;
```
Ideally remove all ternaries though as they make me sad :P

jonathanKingston: Looks like these strings are used in multiple places too, please make it a function if this needs to be the case.

js/lightbeam.js
```diff
@@ -20,6 +24,72 @@ const lightbeam = {
...
+          : `${this.numThirdParties} Third Party Sites`;
+    }
+  },
+
+  async getDataGatheredSince() {
+    const firstRequestUnixTime = await storeChild.getFirstRequestTime();
+    if (!firstRequestUnixTime) {
+      return '';
```
jonathanKingston: Please make this the same as the default like null or undefined even.

js/lightbeam.js
princiya: @biancadanforth can you also write function for each variable to be updated?

js/lightbeam.js
biancadanforth: Okay that's a good call. Done.

js/lightbeam.js
biancadanforth: Done

js/lightbeam.js
biancadanforth: Okay I made a new method `setPartyVar` for this redundancy, and I removed the ternary operators just for you. :)

biancadanforth:
@jonathanKingston Made the changes requested. Ready for another look-see. :)

js/lightbeam.js
```diff
@@ -20,6 +24,82 @@ const lightbeam = {
...
+  // Called from init() (isFirstParty = undefined)
+  // and redraw() (isFirstParty = true or false).
+  async updateVars(isFirstParty) {
+
+    // initialize dynamic vars from storage
+    if (!this.dataGatheredSince) {
+      const { dateStr, fullDateTime } = await this.getDataGatheredSince();
+      this.dataGatheredSince = dateStr;
+      const dataGatheredSince = document.getElementById('data-gathered-since');
+      dataGatheredSince.textContent = this.dataGatheredSince;
```
princiya: @biancadanforth same variable name `dataGatheredSince` is a bit confusing. can we rename the second one to `dataGatheredSinceElement` or maybe `this.dataGatheredSince` to `this.dateStr`?

js/lightbeam.js
```diff
@@ -20,6 +24,82 @@ const lightbeam = {
...
+      dataGatheredSince.setAttribute('datetime', fullDateTime);
+    }
+    if (isFirstParty === undefined) {
+      this.numFirstParties = await this.getNumFirstParties();
+      this.setPartyVar(true);
+      this.numThirdParties = await this.getNumThirdParties();
```
princiya: to make this more readable, can we write this is `this.setPartyVar({ firstParty: true })` or just `this.setPartyVar({ firstParty })` and the next call the thirdParty as `this.setPartyVar({ thirdParty })`?

js/lightbeam.js
biancadanforth: Sure; would be good to differentiate -- done!

js/lightbeam.js
```diff
@@ -20,6 +24,82 @@ const lightbeam = {
...
+        numFirstParties.textContent = `${this.numFirstParties} Sites`;
+      }
+    } else if (this.numThirdParties === 0) {
+      numThirdParties.textContent = '';
+    } else {
+      const str = `${this.numThirdParties} Third Party Sites`;
+      numThirdParties.textContent = str;
+    }
+  },
+
```
princiya: there is no case for `else if` no after `if(isFirstParty)`?
it is only if & else.

js/lightbeam.js
princiya: you can ignore this comment, but can we change `numFirstParties` to `firstPartyCount` ? :)

js/lightbeam.js
princiya: @biancadanforth also, can we only return the number for tp/fp from here?
Example: You have visited - sites & You are connected with - third party sites. That way, the text can stay in the html page and only the number can be returned from here. we could also avoid the whole `if` `else` condition.

js/lightbeam.js
biancadanforth: I replaced the boolean with a string that says what party it is. Couldn't get the objects to work.

index.html
```diff
@@ -43,23 +43,20 @@
...
+              <time id="data-gathered-since" datetime=""></time>
             </dd>
           </div>
           <div>
             <dt>You Have Visited</dt>
-            <!-- @todo get num-sites-visited from store -->
             <dd>
-              <var id="num-sites-visited">6</var> Sites
+              <var id="num-first-parties"></var>
             </dd>
```
princiya:
```
<var><span id="num-first-parties"></span> sites</var>
```

js/lightbeam.js
biancadanforth: I had written it differently, but eslint was throwing an error for `no-lonely-ifs`.

screen shot 2017-08-24 at 11 53 18 am

I too think this is silly, so if you prefer I remove that lint rule, let me know!

biancadanforth: > @biancadanforth also, can we only return the number for tp/fp from here?
@jonathanKingston Asked me to leave the fields blank in HTML until we have a non-zero value, so take that up with him!

index.html
biancadanforth: Again, I initially had this, and @jonathanKingston asked me to change it.

js/lightbeam.js
princiya: eslint is right :) I was thinking in terms of if true, else false. hehe.
i was only reading the code and didn't apply much thought. I am sorry, what you had written initially was fine.
I was seeing a pattern the way fp/tp count was set and was thinking of putting it into a function :). anyways, let me know what you think of the below comment about only returning numeric value from here and setting the text in the html page.

index.html
princiya: ok. @jonathanKingston curious to know why? maybe `<span>` isn't apt inside `<var>`?

biancadanforth: No we just don't think the string "sites" should be there when there's no data (and no value).

princiya: but there will always be 'You have visited' text. So why not return 0 when there is no data?

biancadanforth: Because we don't have a date value until they visit a site, so we'd have inconsistent entries (no value for date, a zero value for the other two). Honestly, I don't care either way, just saying I had it the way you suggest before, and I'm more inclined this way. We'll let @jonathanKingston decide I suppose!

princiya: cool no problem. i had missed this discussion :)

biancadanforth:
It's easy enough to change, either way.

index.html
jonathanKingston: Just that my understanding of the semantics are that var tags should just have data rather than sentences etc.

jonathanKingston: https://developer.mozilla.org/en-US/docs/Web/HTML/Element/var << represents a variable in the mathematical sense. It's either wrong to use at all or better being how we have it.

Issue #66, opened 2017-06-19
Fix toggle button to work in CSS


jonathanKingston:
Remove the hard coding of "off" and use CSS to control the toggle button.

The functionality of this button can be done elsewhere.

jonathanKingston:
**Note** the functionality itself depends on a web extension API that doesn't exist yet. See: https://bugzilla.mozilla.org/show_bug.cgi?id=1345158

laxmy:
Hi Lightbeam Team,
I'm interested in working/contributing on this project. 
This looks like something I can take up. Is this up for grabs?

jonathanKingston:
@laxmy the initial work here is to duplicate a toggle that is available for the current Lightbeam (top right of the page) using CSS and semantic markup as much as possible (so using a native checkbox and style correclty based on `:checked` status etc)

The timescale is a little tight here so we might have to step in if we need to go faster, however yeah we can take patches for any "good first bug".

I'm going to assign this to you for the time being. Shout if you have questions etc.

@biancadanforth as a heads up this is off your plate however when you get time from p1's this could use the backend to actually enable the tracking protection toggle.

biancadanforth:
@laxmy Thanks for you interest in helping out! We already have some HTML/CSS for this toggle, the markup is copied below. You can view the toggle by removing the `unimplemented` class on the `div.tracking-protection` element. Feel free to use this as a point of departure and reference regarding the color scheme.

```html
<!-- @todo add tracking protection and toggle functionality -->
  <div class="tracking-protection unimplemented">
    <label for="tracking-protection-control">Tracking Protection</label>
    <div class="toggle-switch">
      <!-- @todo couple checkbox status (checked/unchecked) with toggle status -->
      <input type="checkbox" id="tracking-protection-control">
      <span class="toggle"></span>
      <!-- @todo toggle-switch text (off/on) to reflect state of tracking protection -->
      <span class="toggle-status">Off</span>
  </div>
</div>
```

jonathanKingston:
Ideally it won't need toggle or toggle status and could be done with CSS `::before` and `::after` instead but it may not be workable.

laxmy:
@jonathanKingston and @biancadanforth ....Thanks for the info and the starting point. I'm working on this. The alert window that shows info about tracking protection will be enabled from the back-end is my understanding.

jonathanKingston:
This is fixed by the latest merge from @laxmy thanks again.

PR #170, opened 2017-08-23
#66 Changes proposed for getting the toggle button up

laxmy:
Here are the changes I did to get the toggle button up. 
I have done it using css :before and :after to 'toggle'  like @jonathanKingston proposed to get the button and the 'on'-'off' label. I have also made changes to the button dimensions to make it look more like the original lightbeam switch.
Hope I used the right color.
Let me know if this works.
#66 Fix toggle button to work with CSS

### DIFF
--- a/css/style.css
+++ b/css/style.css
@@ -224,10 +224,10 @@ main {
   align-items: center;
   justify-content: space-around;
   height: 30px;
-  width: 60px;
+  width: 55px;
   margin-left: 5px;
   background-color: var(--secondary-color);
-  border-radius: 2px;
+  border-radius: 5px;
   position: relative;
 }
 
@@ -243,7 +243,7 @@ main {
   border-radius: 5px;
   transition:all 0.4s;
   position: absolute;
-  left:10%;
+  left:5px;
   top:10%;
 }
 
@@ -259,7 +259,7 @@ main {
 }
 
 input:checked + .toggle::before{
- transform: translateX(22px);
+ transform: translateX(20px);
  right:10%;
  background-color: var(--secondary-text-color);
 }
--- a/css/style.css
+++ b/css/style.css
@@ -224,8 +224,8 @@ main {
   align-items: center;
   justify-content: space-around;
   height: 30px;
-  width: 70px;
-  margin-left: 10px;
+  width: 60px;
+  margin-left: 5px;
   background-color: var(--secondary-color);
   border-radius: 2px;
   position: relative;
@@ -237,14 +237,14 @@ main {
 
 .toggle::before {
   content: "";
-  height: 20px;
-  width: 30px;
+  height: 24px;
+  width: 25px;
   background-color: var(--primary-color);
   border-radius: 5px;
   transition:all 0.4s;
   position: absolute;
   left:10%;
-  top:20%;
+  top:10%;
 }
 
 .toggle::after {
@@ -259,7 +259,7 @@ main {
 }
 
 input:checked + .toggle::before{
- transform: translateX(25px);
+ transform: translateX(22px);
  right:10%;
  background-color: var(--secondary-text-color);
 }
--- a/css/style.css
+++ b/css/style.css
@@ -220,7 +220,7 @@ main {
 }
 
 .toggle-switch {
-  display: flex;
+  display: inline-flex;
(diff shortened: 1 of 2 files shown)


css/style.css
```diff
@@ -220,34 +220,52 @@ main {
...
-  border-radius: 2px;
+  border-radius: 5px;
+  position: relative;
 }
 
-input[type="checkbox"] {
-  -moz-appearance: none;
-  margin: 0;
+.toggle-switch input{
+  display: none;
```
jonathanKingston: Can you use opacity here and `-moz-appearance: none;` then check that tabbing to the element and hitting space bar activates the element (it won't being display none).

jonathanKingston: This actually might allow you to remove the `span class="toggle"` also if you make the input visible but slide along?

index.html
```diff
@@ -65,16 +65,15 @@
           </div>
         </dl>
         <!-- @todo add tracking protection and toggle functionality -->
-        <div class="tracking-protection unimplemented">
+        <div class="tracking-protection">
           <label for="tracking-protection-control">Tracking Protection</label>
+          <label>
```
jonathanKingston: Can this use the label above instead of using a new element here?
Something like:
```
<div class="tracking-protection">
  <label for="tracking-protection-control">
    Tracking Protection
      <div class="toggle-switch">
        <input type="checkbox" id="tracking-protection-control">
      </div>
    </label>
</div>
```
Indenting seems wrong anyway.

css/style.css
laxmy: Tried the accessibility fix. But I'm unable to tab the toggle-switch and activate it. The main area as a whole gets tab focus. Not sure why.
`.toggle-switch {
  display: inline-flex;
  align-items: center;
  justify-content: space-around;
  height: 30px;
  width: 55px;
  margin-left: 5px;
  background-color: var(--secondary-color);
  border-radius: 5px;
  position: relative;
  cursor: pointer;
}

.toggle-switch input {
  -moz-appearance: none;
  opacity: 0;
}

.toggle-switch:active {
  border:1px solid var(--button-active-border-color);
}`

jonathanKingston: If you make the opacity not 0 can you see it? Style it like the switch instead?

Also I think the active is on the wrong element: `.toggle-switch:active` instead should be: `.toggle-switch input:active`

Something like:
```
.toggle-switch input {
  -moz-appearance: none;
}

.toggle-switch input:checked::before {
  ...
}

.toggle-switch input::after {
  ...
}

.toggle-switch input:checked {
  ...
}

.toggle-switch input:checked::after {
  ...
}

.toggle-switch input:active {
  border:1px solid var(--button-active-border-color);
}
```

css/style.css
```diff
@@ -220,34 +220,52 @@ main {
...
+  position: absolute;
+  right: 10%;
+  top:30%;
+}
+
+input:checked + .toggle::before{
+ transform: translateX(20px);
+ right:10%;
+ background-color: var(--secondary-text-color);
+}
```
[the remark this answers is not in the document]
laxmy: Done indenting changes in the last commit

css/style.css
```diff
@@ -220,34 +220,52 @@ main {
...
   background-color: var(--secondary-color);
-  border-radius: 2px;
+  border-radius: 5px;
+  position: relative;
 }
 
-input[type="checkbox"] {
-  -moz-appearance: none;
-  margin: 0;
+.toggle-switch input{
```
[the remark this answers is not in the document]
laxmy: Done adding the cursor and active state

laxmy:
@jonathanKingston : I made the indenting changes, active state , cursor and accessibility fix you recommended. But I needed the extra label and span elements to contain the switch. The label is what holds the checked state of checkbox . Styling checkbox itself like a switch did not work. Also, as checkbox with opacity 0 makes it child elements also transparent including :before and :after I had to add the extra span element to hold the :before and :after elements. Hope this works!

jonathanKingston:
Thanks a lot @laxmy ! I merged this and will hook up the code to make it work. Thanks!

laxmy:
@jonathanKingston: Thanks for merging my changes and the guidance!

PR #173, opened 2017-08-23
[README.md] adding separators in between topics

tiagmoraismorgado:
---

**this PR does basically aim at:**
- *adding separators in between topics in the README.md file;*

---

**main motivation for that does basically lie within the fact within:**
- *turning the file simpler to to be read;*
- *turning the file easier to read;*
- *turning the file easier to be understood;*
- *turning the file simpler to to be understood;*
- *stylizing/beautifying the file;*

---

### DIFF
--- a/README.md
+++ b/README.md
@@ -3,6 +3,8 @@ This is the web extension version of the Mozilla Lightbeam add-on for visualizin
 
 The Mozilla Lightbeam extension is a key tool for Mozilla to educate the public about privacy
 
+---
+
 ## Quick Start
 
 ### Clone the repository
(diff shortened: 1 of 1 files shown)


jonathanKingston:
Given the default styles on Github are to add an underline on headings this actually makes headings appear like this:

---
# heading

This for me at least makes it harder to read on here which will be the highest place the README will be consumed.
This could be a bug for later perhaps when we have a longer README but certainly at the moment I think this is overkill.

Thanks for the patch, at the moment I don't think we need it so I'm going to close.

Issue #110, opened 2017-07-26
Consider how to draw first parties that are also third parties


biancadanforth:
For example: www.google.com is a first party site, but is also a third party site for www.youtube.com. If you visit www.google.com and then www.youtube.com, the google first party node disappears in the visualization currently, because the `www.google.com` website object gets assigned an `isVisible` key with a value of `false` once is it linked as a third party by www.youtube.com.

When a website object has an `isVisible` key set to `false`, it is ignored by the `getAll` method currently, and likely also by the per-website updates, but this should be confirmed.

In general, we should consider how these types of sites, that are both first and third parties, are handled and show up in the visualization in both the case of a page refresh (a call to `store.getAll`) and an update (`store.updateChild`) inside `store.setFirstParty` and `store.setThirdParty`.

@jonathanKingston I am guessing this is P1, but please update the priority flag as you deem appropriate!

biancadanforth:
@jonathanKingston Found another symptom of this bug:

![firstpartiesalsothirdpartiesbug](https://user-images.githubusercontent.com/17437436/28724231-2c9e0786-736e-11e7-8f76-8bc8a78c06df.png)

**What we believe is happening:**
* A first party site (Site A) is initially drawn as expected to the visualization, linking to a third party, Site B.
* Later, another website (Site C) links to Site A as a third party, and in this case Site A is overwritten as a third party and is whitelisted.
* This gives Site A an `isVisible` key set to `false`
* When the visualization gets updated, Site A gets removed from the graph, but the third party, Site B that it originally linked to remains, without the link.

biancadanforth:
@jonathanKingston Can you provide more details about how this could be tested? You had mentioned something about local servers in our weekly meeting which I didn't follow...

My thought:
Just as we have an allowlist override for third parties owned by first parties, perhaps we have an allowlist override override if that third party is already in storage listed as a first party?

I.e. for Google.com, if it is already drawn on the visualization as a first party and is then linked to as a third party, just leave it alone (don't make its `isVisible` key `false`) and add the new link.

jonathanKingston:
> for Google.com, if it is already drawn on the visualization as a first party and is then linked to as a third party

This is clear for google.com however more obscure sites would need to be elevated from being a third party to being a first party. Treating them consistently to appear as a first party would be the correct behavior here.
I agree that how current lightbeam does this is what we want also.

jonathanKingston:
Visit:
[...]

PR #175, opened 2017-08-24
[WIP] Prevent ghosting - P1

princiya:
Issue #147

### DIFF
--- a/js/viz.js
+++ b/js/viz.js
@@ -3,6 +3,7 @@ const viz = {
   scalingFactor: 2,
   circleRadius: 5,
   resizeTimer: null,
+  d3Timer: null,
   minZoom: 0.5,
   maxZoom: 1.5,
   collisionRadius: 30,
@@ -353,7 +354,7 @@ const viz = {
     d3.event.subject.fy = d3.event.y;
 
     this.hideTooltip();
-    this.drawOnCanvas();
+    this.d3Timer = d3.timer(() => this.drawOnCanvas());
   },
 
   dragEnd() {
@@ -365,6 +366,7 @@ const viz = {
     d3.event.subject.fx = null;
     d3.event.subject.fy = null;
     d3.event.subject.shadow = false;
+    this.d3Timer.stop();
   },
 
   addZoom() {


princiya:
@jonathanKingston I have used `d3.timer` which internally uses `requestAnimationFrame`.
There isn't ghosting now, but the repaint is too slow.

There is a related D3 issue to improve performace using web worker - [https://github.com/d3/d3/issues/1519](https://github.com/d3/d3/issues/1519).

Reading a bit and trying that out!

princiya:
As I was debugging, I realised in `lightbeam.js` we have the `redraw()` function which calls `viz.draw()`.
This redraw function passes the whole `this.websites` object to the `simulation`.

Each time we call `viz.draw()` we are repainting the whole graph and not letting D3 decide which are the new nodes added dynamically. Maybe this is the reason, I couldn't get `tick` to work because the graph never stabilised as it grew, as it was trying to repaint the whole graph and not account to just the newly added nodes.

When we first started with D3, we had D3's join methods - `enter()`&`exit()`.

[This](https://bl.ocks.org/mbostock/ad70335eeef6d167bc36fd3c04378048) example graph has a fair amount of nodes and is a force layout example using `canvas`. This fits our use case, except there is no dynamic behaviour. And there is no ghosting issue during `drag`.

So, I shall try and follow the below approach:
- pass only new nodes from `lightbeam.js` into `viz.draw()`
- use D3's join methods to take into account for the dynamically added new nodes
- don't use manual tick, instead use D3's `tick` handler

biancadanforth:
@princiya Thanks for writing your thoughts out for the rest of us!

My understanding of D3 is that when we use the `join()`, `enter()` and `exit()` methods, it is figuring out which nodes to keep the same, add and remove, respectively, so that we don't have to write that logic ourselves in lightbeam.js. When I was messing around with D3 initially using these methods, my guess was that the resetting issue was not that we were passing all the website data every time in, but that the simulation and forces were resetting every time.

Here is an example and a good resource on these methods that I was looking at way back when (you probably have seen them before, but just in case):
* [General Update Pattern](https://bl.ocks.org/mbostock/3808218)
  - This passes in a randomly generated string of characters every 1.5 seconds and D3 figures out which characters to add, remove and update (update the x,y positions of).
* [Thinking with Joins](https://bost.ocks.org/mike/join/)

Now that we are no longer passing in a `websites` blob to Lightbeam, it may not be too difficult to update one node at a time, as you suggest, but it would be great to know in general if D3 handles this for us with these methods (I didn't realize we were no longer using them).

Either way I'm sure you'll get it working with the `tick` event!

biancadanforth:
As you pointed out in slack, these `join`, `enter` and `exit` methods require DOM selections -- that's probably why all the [examples](https://bocoup.com/blog/d3js-and-canvas/) you found when you were just transitioning the visualization from svg to canvas use DOM objects still. Would doing this be a terrible idea @princiya @jonathanKingston ?

The example above uses `canvas` with the `join`, `enter` and `exit` methods.

princiya:
@biancadanforth we are still good with the 'objects in memory' approach :) 

As far as join methods are concerned, the `redraw()` function gets the updated data (fp/tp) and we have to pass only this piece to the `viz.draw()` instead of the whole `websites` object. We wouldn't need `enter()` or `exit()` methods from D3 join. We don't have a case for `exit()` anyways, that rules it out completely.

biancadanforth:
Thanks Princiya! Though it should be noted that there are updates where we are not adding a new shape/line to the canvas -- the only example I can think of right now is when a third party site later is linked to as a first party site (#110)  -- in this case we want to replace the triangle at that node's location with a circle -- is that possible with this approach?

princiya:
@biancadanforth Thanks for the above pointer. At the moment, I do not have an answer, but I will keep this in mind.

jonathanKingston:
@princiya I think we can close this right? I don't see this anymore with your latest fixes.

princiya:
@jonathanKingston closing this.

PR #176, opened 2017-08-25
Address first parties as third parties. Fixes #110.

### LINKED ISSUE #110: Consider how to draw first parties that are also third parties

biancadanforth:
If a first party site later becomes a third party site, keep the firstParty key set to true and do not set isVisible to false, even if the site is on the allowlist as a third party. Added a new method 'isAlsoAThirdParty' to check this.

Before:
screen shot 2017-08-25 at 12 44 20 pm

After:
screen shot 2017-08-25 at 1 49 35 pm

@jonathanKingston , I have no known questions; so ready for you to take a look. :)

### DIFF
--- a/js/store.js
+++ b/js/store.js
@@ -296,7 +296,8 @@ const store = {
     // and the third party is visible (i.e. not allowlisted)
     if (!this.isFirstPartyLinkedToThirdParty(firstParty, target)) {
       if (!this.isVisibleThirdParty(thirdParty)) {
-        if (this.onAllowList(origin, target)) {
+        if (this.onAllowList(origin, target)
+          && !this.isAlsoAFirstParty(thirdParty)) {
           // hide third party
           thirdParty['isVisible'] = false;
         } else {
@@ -320,6 +321,10 @@ const store = {
     // merge data with thirdParty
     // @todo restructure storage to use data from capture for graph filtering
     for (const key in data) {
+      if (key ==='firstParty' && this.isAlsoAFirstParty(thirdParty)) {
+        // third parties that are already a first party stay a first party
+        data[key] = true;
+      }
       thirdParty[key] = data[key];
     }
 
@@ -343,6 +348,10 @@ const store = {
       return false;
     }
     return true;
+  },
+
+  isAlsoAFirstParty(thirdParty) {
+    return thirdParty.firstParty;
   },
 
   async addFirstPartyLink(firstPartyHostname, thirdPartyHostname) {


js/store.js
```diff
@@ -320,6 +321,10 @@ const store = {
     // merge data with thirdParty
     // @todo restructure storage to use data from capture for graph filtering
     for (const key in data) {
+      if (key ==='firstParty' && this.isAlsoAFirstParty(thirdParty)) {
```
jonathanKingston: Nit space after `===`

js/store.js
```diff
@@ -345,6 +350,10 @@ const store = {
     return true;
   },
 
+  isAlsoAFirstParty(thirdParty) {
```
jonathanKingston: Should this function work on both fp+tp's instead of just being given a node. This could check both keys.

js/store.js
biancadanforth: Fixed, thank you... is there an eslint rule for that?

js/store.js
biancadanforth: If I am understanding you correctly... The default behavior is to overwrite the value of `firstPaty` (whether `true` or `false`) with the new request data, so we only need to worry about the case when a first party later becomes a third party (we want to preserve `firstParty` as `true`), and that's what this method checks.

I am open to a less confusing way of doing this!

js/store.js
jonathanKingston: Probs consistent whitespace one. Feel free to file a bug, right now don't have time to check this.

js/store.js
jonathanKingston: I think we need to cope when a either type become a dual. So I think this check should happen for both on update.

js/store.js
biancadanforth: Okay I found it; it's `space-infix-ops`. I added it to `eslintrc.json`.

biancadanforth:
Superceded by PR #180 .

PR #178, opened 2017-08-29
Refactor store.setThirdParty

biancadanforth:
Removed the nested `if` statements for clarity. This may help in troubleshooting #110.

Also added an eslint rule to ensure there's a space on either side of operators so we don't get things like:

```javascript
if (someVar ===4) {
  // do stuff
}
```

### DIFF
--- a/js/store.js
+++ b/js/store.js
@@ -278,7 +278,6 @@ const store = {
       throw new Error('setThirdParty requires a valid origin argument');
     }
 
-    let isNewThirdParty = false;
     let shouldUpdate = false;
 
     const firstParty = await this.getWebsite(origin);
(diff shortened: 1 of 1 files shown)


jonathanKingston:
This has introduced some linting issues in viz.js which I was going to raise at the time :).

I also need to double check the logic changes as it's not just a refactor, this is also part of the most complex code in Lightbeam so I'm nervous to merge right now.

js/store.js
```diff
@@ -292,29 +291,23 @@ const store = {
...
-          shouldUpdate = true;
-        }
-      }
-      if (this.isVisibleThirdParty(thirdParty) && !isNewThirdParty) {
-        // an existing visible third party links to a new first party
-        await this.addFirstPartyLink(origin, target);
-        shouldUpdate = true;
+    // isVisible only exists on an allowlisted
+    // or previously allowlisted third party
+    if (!thirdParty['isVisible'] && !this.onAllowList(origin, target)) {
```
jonathanKingston: This should likely have a newline before `&&` for consistency.

js/store.js
```diff
@@ -292,29 +291,23 @@ const store = {
       thirdParty['firstPartyHostnames'].push(origin);
     }
 
-    // add link in first party if it doesn't exist yet
-    // and the third party is visible (i.e. not allowlisted)
-    if (!this.isFirstPartyLinkedToThirdParty(firstParty, target)) {
-      if (!this.isVisibleThirdParty(thirdParty)) {
```
jonathanKingston: I preferred this method for readability, removing it might change the logic what is the rationale? Is it still used?

js/store.js
```diff
@@ -292,29 +291,23 @@ const store = {
       thirdParty['firstPartyHostnames'].push(origin);
     }
 
-    // add link in first party if it doesn't exist yet
-    // and the third party is visible (i.e. not allowlisted)
-    if (!this.isFirstPartyLinkedToThirdParty(firstParty, target)) {
```
jonathanKingston: `isFirstPartyLinkedToThirdParty` has been removed on the ` if (!thirdParty['isVisible'] && !this.onAllowList(origin, target)) ` section of your code, won't this make updates overfire?

js/store.js
jonathanKingston: This previously was in place to check for missing `isVisible` keys. Perhaps this is no longer needed though.

jonathanKingston:
It looks like `isVisible = false` is never set, which from my initial checking looks like it would never be treated as having an allowList at all.

biancadanforth:
Let's leave the logic as is for now; I'll revisit this time permitting, otherwise you can close it. :)

jonathanKingston:
I didn't actually load it, I just dunno how it would ever be a not visible node from the code I looked at.
I can leave it open however I worry about changing this.

biancadanforth:
Going to close this PR for now, as it's not critical. We can always revisit later.

Issue #179, opened 2017-08-29
Some allowlist third parties show up on the graph


biancadanforth:
See [this comment](https://github.com/mozilla/lightbeam-we/issues/110#issuecomment-325550561) for an explanation.

I believe this is a related issue to #110 .

biancadanforth:
Basically, `youtube` starts out as a first party, and later becomes an allowlisted third party (so that it's `firstParty` key is set to `false` and `isVisible` key is also set to `false`). This means when we refresh the page, `getAll`, which only returns nodes that are either `firstParty` or `isVisible` `true`, ignores `youtube`, so any nodes it was originally connected to in the visualization lose their `youtube` link.

In general, the logic of `setThirdParty` should be such that `firstParty` and `isVisible` should **never** be changed if they are ever `true`.

jonathanKingston:
Should be fixed with latest patch.

Issue #177, opened 2017-08-28
Third->third line issues


jonathanKingston:
I'm seeing when loading:
- https://www.wikipedia.org/
- searching sdsd
- getting: https://en.wikipedia.org/wiki/Special:Search?search=sdsd&go=Go&searchToken=e4s8zqqpwutej8y7gvkerdca6
- clicking first result: https://en.wikipedia.org/wiki/South_Dakota_School_for_the_Deaf

This gets the following:
![image](https://user-images.githubusercontent.com/338988/29796638-25b9892a-8c07-11e7-94ed-1dbf52d57cc4.png)

```
{
  "en.wikibooks.org": {
    "hostname": "en.wikibooks.org",
    "favicon": "",
    "firstPartyHostnames": [
      "en.wikipedia.org"
    ],
    "firstParty": false,
    "thirdParties": []
  },
  "en.wikipedia.org": {
    "hostname": "en.wikipedia.org",
    "favicon": "https://en.wikipedia.org/static/favicon/wikipedia.ico",
    "firstPartyHostnames": [
      "www.wikipedia.org",
      "en.wikipedia.org"
    ],
    "firstParty": false,
    "thirdParties": []
  },
  "login.wikimedia.org": {
    "hostname": "login.wikimedia.org",
    "favicon": "",
    "firstPartyHostnames": [
      "en.wikipedia.org"
    ],
    "firstParty": false,
    "thirdParties": []
  },
  "meta.wikimedia.org": {
    "hostname": "meta.wikimedia.org",
    "favicon": "",
    "firstPartyHostnames": [
      "en.wikipedia.org"
    ],
    "firstParty": false,
    "thirdParties": []
  },
  "upload.wikimedia.org": {
    "hostname": "upload.wikimedia.org",
    "favicon": "",
    "firstPartyHostnames": [
      "www.wikipedia.org",
      "en.wikipedia.org"
    ],
    "firstParty": false,
    "thirdParties": []
  },
  "www.wikipedia.org": {
    "hostname": "www.wikipedia.org",
    "favicon": "https://www.wikipedia.org/static/favicon/wikipedia.ico",
    "firstPartyHostnames": [
      "www.wikipedia.org"
    ],
    "firstParty": false,
    "thirdParties": []
  }
}
```

So for some reason we are seeing third parties having third parties which isn't expected. We then are not rendering third->third parties which maybe we should do anyway (not that the data should ever do this - this could be part of the ft+tp fix).

[...]

PR #180, opened 2017-08-29
Treat data with more consistency #177 possibly #179 and issues with #110

jonathanKingston:
Changes include:
- Ensuring modified data that is stored gets passed back in update calls
- Inbound, outbound data munging to be the same shape (I think there are some bits missing to this)
- Ensures we draw the graph based on the thirdParties and firstParties keys rather than what type the node is
- Merges data for isVisible and firstParty so they never go from true to false after being set
- Fixes race condition where tp fires before fp causing a broken fp node.

### DIFF
--- a/js/lightbeam.js
+++ b/js/lightbeam.js
@@ -189,7 +189,7 @@ const lightbeam = {
     let links = [];
     for (const website in this.websites) {
       const site = this.websites[website];
-      if (site.firstParty) {
+      if (site.thirdParties) {
         const thirdPartyLinks = site.thirdParties.map((thirdParty) => {
           return {
             source: website,
@@ -236,7 +236,7 @@ const lightbeam = {
       this.websites[data.hostname] = data;
       this.updateVars(data.firstParty);
     }
-    if (!data.firstParty) {
+    if (data.firstPartyHostnames) {
       // if we have the first parties make the link if they don't exist
       data.firstPartyHostnames.forEach((firstPartyHostname) => {
         if (this.websites[firstPartyHostname]) {
--- a/js/store.js
+++ b/js/store.js
@@ -10,13 +10,18 @@ const store = {
     await this.getAllowList();
   },
 
+  indexes: [
+    'hostname', // Primary key
+    'firstRequestTime',
+    'lastRequestTime',
+    'isVisible',
+    'firstParty'
+  ],
+
   makeNewDatabase() {
     this.db = new Dexie('websites_database');
-    this.db.version(1).stores({
-      // store: 'primaryKey, index1, index2, ...'
-      websites:
-        'hostname, firstRequestTime, lastRequestTime, isVisible, firstParty'
-    });
+    const websites = this.indexes.join(', ');
+    this.db.version(1).stores({websites});
     this.db.open();
   },
 
@@ -144,16 +149,11 @@ const store = {
       hostname: hostname,
       favicon: website.faviconUrl || '',
       firstPartyHostnames: website.firstPartyHostnames || false,
-      firstParty: 0,
+      firstParty: !!website.firstParty,
       thirdParties: []
     };
-    if (website.firstPartyHostnames === undefined) {
-      output.firstParty = 1;
-      if ('thirdPartyHostnames' in website) {
-        output.thirdParties = website.thirdPartyHostnames;
-      } else {
-        output.thirdParties = [];
-      }
+    if ('thirdPartyHostnames' in website) {
+      output.thirdParties = website.thirdPartyHostnames;
     }
     return output;
   },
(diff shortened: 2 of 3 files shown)


biancadanforth:
I see that youtube and google are first parties now even after refresh, so that's good. But this still suffers from at least a couple symptoms:

Going to:
1. https://www.google.com/search?q=dfdfdf&ie=utf-8&oe=utf-8&client=firefox-b-ab
2. https://www.youtube.com/
3. https://developers.google.com/youtube/youtube_player_demo

`www.youtube.com` still links to allowlisted third parties (depending on the order of requests; this is intermittent):
- s1.2mdn.net
- securepubads.g.doubleclick.net

Also refreshing the page still has the same issue of showing a completely different graph.

fpandtpbug

jonathanKingston:
Test it again the latest update fixes the issue you mentioned.

jonathanKingston:
Changes include:
- Ensuring stored data is passed to updates
- Inbound, outbound data munging to be the same shape (I think there are some bits missing to this)
- Ensures we draw the graph based on the thirdParties and firstParties keys rather than what type the node is
- Merges data for isVisible and firstParty so they never go from true to false after being set
- Fixes race condition where tp fires before fp causing a broken fp node.

jonathanKingston:
There may be a race condition with populating the allow list perhaps because I can't replicate what you are saying.

I think this supercedes the pr for #110

jonathanKingston:
First parties count sometimes doesn't populate for me, I will check if this causes it.

biancadanforth:
I don't see the capturing crashing, but I am definitely seeing that the dynamic vars aren't matching the visualization **on refresh** (says 'you have visited' 1 site and should be more like 10):

screen shot 2017-08-29 at 3 26 11 pm

This happens also on the `master` branch, but to a lesser extent for these same sites (i.e. says 9 instead of 10, and i think the count is only off by one because youtube gets swapped to a third party on reload on the master branch).

js/store.js
```diff
@@ -144,16 +149,11 @@ const store = {
       hostname: hostname,
       favicon: website.faviconUrl || '',
       firstPartyHostnames: website.firstPartyHostnames || false,
-      firstParty: 0,
+      firstParty: !!website.firstParty,
```
biancadanforth: `website.firstParty` is coming in as 1 and 0 instead of true and false due to `setWebsite` now "munging" the data, so you use the `!!` to coerce the value into a boolean data type in `outputWebsite` for the visualization.

js/store.js
```diff
@@ -195,42 +195,85 @@ const store = {
...
+        case 'requestTime':
+          // store first and last request times for clearing data every X days
+          if (!('firstRequestTime' in website)) {
+            website.firstRequestTime = value;
+          } else {
+            website.lastRequestTime = value;
+          }
+          break;
+        case 'isVisible':
+          if ('isVisible' in website
```
biancadanforth: This logic is saying we NEVER want to change `isVisible` to `false` if it is already set to `true`.

js/store.js
```diff
@@ -195,42 +195,85 @@ const store = {
...
+              && website.isVisible === true) {
+            break;
+          }
+          website.isVisible = value;
+          break;
+        case 'firstParty':
+          if ('firstParty' in website
+              && website.firstParty === true) {
```
biancadanforth: This is saying we NEVER want a site that's already stored as a `firstParty` `true` to be switched to `false`.

js/store.js
```diff
@@ -366,6 +409,10 @@ const store = {
 
   async addFirstPartyLink(firstPartyHostname, thirdPartyHostname) {
     const firstParty = await this.getWebsite(firstPartyHostname);
+    // We are likely storing the tp first, lets account for that
```
biancadanforth: Ah, because the `setFirstParty` doesn't write the first party in the database until after the tab has loaded (for visible tabs), we may get an empty object for a while from storage.

Since we call `setFirstParty` from `setThirdParty`, we could store this information sooner than waiting for `setFirstParty` to execute?

js/lightbeam.js
```diff
@@ -189,7 +189,7 @@ const lightbeam = {
     let links = [];
     for (const website in this.websites) {
       const site = this.websites[website];
-      if (site.firstParty) {
+      if (site.thirdParties) {
```
biancadanforth: This is saying: We are actually just interested to know what the new links are for a site, which could be both a first and a third party, so let's check for that instead.

js/lightbeam.js
```diff
@@ -236,7 +236,7 @@ const lightbeam = {
       this.websites[data.hostname] = data;
       this.updateVars(data.firstParty);
     }
-    if (!data.firstParty) {
+    if (data.firstPartyHostnames) {
```
biancadanforth: Same goes for here. We just want to know what the new links are, if any and the `firstParty` boolean doesn't reflect that.

js/store.js
```diff
@@ -144,16 +149,11 @@ const store = {
...
       thirdParties: []
     };
-    if (website.firstPartyHostnames === undefined) {
-      output.firstParty = 1;
-      if ('thirdPartyHostnames' in website) {
-        output.thirdParties = website.thirdPartyHostnames;
-      } else {
-        output.thirdParties = [];
-      }
+    if ('thirdPartyHostnames' in website) {
```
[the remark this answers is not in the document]
jonathanKingston: Yeah, **todo** I need to check that we never send firstPartyHostname updates when the node is not visible.

js/store.js
jonathanKingston: Yep, update was getting called with no `firstParty` key which made the graph broken on update view.

jonathanKingston:
Yeah just one thing to double check because I am paranoid. Will look when home.

jonathanKingston:
Going to merge this as agreement with @biancadanforth that this solves many of the issues involved. I'm sure some might still exist to a smaller scale.

PR #181, opened 2017-08-29
Fix TypeError for a null return value in lightbeam.getDataGatheredSince

biancadanforth:
If storage is empty, just exit the method to update the dynamic vars.


js/lightbeam.js
```diff
@@ -31,6 +31,9 @@ const lightbeam = {
     // initialize dynamic vars from storage
     if (!this.dataGatheredSince) {
       const { dateStr, fullDateTime } = await this.getDataGatheredSince();
+      if (!dateStr) {
```
jonathanKingston: Seems pretty bad form to have such different shaped data, however this will fix the initial issue so will merge.

PR #182, opened 2017-08-29
Fix null error in lightbeam.getDataGatheredSince

biancadanforth:
Follow on to PR #181


biancadanforth:
What do you mean about the shape? I'm more than happy to make changes.

PR #186, opened 2017-08-31
Add first run page. Fixes #117.

### LINKED ISSUE #117: Make a 'first run' page
The legacy Lightbeam has a 'first run' page the first time Firefox is run after installing Lightbeam. This helps to explain to the user what to expect and how to make the most of it.

This is a good idea that should be implemented in this updated version of Lightbeam.

Here's the legacy add-on's first run page as a reference:

lightbeam-first-ff-run

biancadanforth:
We probably want to make a follow-up issue to check for first run with the browser extension version instead of using local storage and add a button in first-run.html to open the Lightbeam tab from the first-run page.


biancadanforth:
@jonathanKingston This is an initial first run page that I think will do the job. Please check and let me know your thoughts on it and follow-up issues for later (see commit message).

PR #188, opened 2017-08-31
Properly munge inbound data in all cases. Fixes #187.

### LINKED ISSUE #187: Fix dynamic vars on refresh
The inbound data to storage is not being properly munged in the `switch` statement for case `'firstParty'` and case `'isVisible'`. This leads to incorrect or no data for how many sites the user has visited in the UI.

See [this comment](https://github.com/mozilla/lightbeam-we/pull/180#issuecomment-325823113) for details.

biancadanforth:
Dynamic vars were not updating correctly with the first party count in the switch statement for the `isVisible` and `firstParty` cases in `store.setWebsite`.

### DIFF
--- a/js/store.js
+++ b/js/store.js
@@ -141,6 +141,9 @@ const store = {
   },
 
   async _write(website) {
+    for (const key in website) {
+      website[key] = this.mungeDataInbound(key, website[key]);
+    }
     return await this.db.websites.put(website);
   },
 
@@ -239,7 +242,7 @@ const store = {
     }
 
     for (const key in data) {
-      const value = this.mungeDataInbound(key, data[key]);
+      const value = data[key];
       switch (key) {
         case 'requestTime':
           // store first and last request times for clearing data every X days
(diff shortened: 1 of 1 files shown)


biancadanforth:
@jonathanKingston , there might be a more elegant way to do this, but this works. :)

js/store.js
```diff
@@ -251,19 +251,23 @@ const store = {
           break;
         case 'isVisible':
           if ('isVisible' in website
-              && website.isVisible === true) {
-            break;
+              && website.isVisible === true
+              && value === 0) {
```
jonathanKingston: I'm lost, isn't this doing exactly the same other than allowing perhaps different values later?

js/store.js
```diff
@@ -251,19 +251,23 @@ const store = {
...
-            break;
+              && website.firstParty === true
+              && value === 0) {
+            // once a website is a first party, it will always be drawn as one
+            value = 1;
           }
-          website.firstParty = value;
+          website[key] = value;
           if (value) {
-            website.isVisible = true;
```
jonathanKingston: This line here is the only cahnge I can see^

biancadanforth:
@jonathanKingston , made changes as we discussed in slack. It's much cleaner now. Let me know what you think!

PR #193, opened 2017-09-01
Hook up tracking protection toggle with web extensions.

jonathanKingston:


groovecoder:
Ooo ... spiffy! What version introduced `browser.privacy.websites.trackingProtectionMode`? Needs to be added to https://developer.mozilla.org/en-US/Add-ons/WebExtensions/API/privacy/websites

jonathanKingston:
@groovecoder I think 57, let me check the bug if it is docs flagged. I don't think it gets on MDN until stable.

jonathanKingston:
@groovecoder yeah has the dev-doc-needed flag: https://bugzilla.mozilla.org/show_bug.cgi?id=1345158

Issue #119, opened 2017-08-08
Provide warning when removing all data


jonathanKingston:
Current Lightbeam has a warning:

![selection_780](https://user-images.githubusercontent.com/338988/29068071-1d5f646e-7c2d-11e7-9a72-65ebd296e066.png)

PR #197, opened 2017-09-03
add warning when remove all data

joanasouza:
Issue #119

### DIFF
--- a/js/lightbeam.js
+++ b/js/lightbeam.js
@@ -142,7 +142,10 @@ const lightbeam = {
   resetData() {
     const resetData = document.getElementById('reset-data-button');
     resetData.addEventListener('click', async () => {
-      await storeChild.reset();
+      var confirmation = confirm('Pressing OK will delete all'
+       + ' Lightbeam information. Are you sure?');
+      if (confirmation == true)
+        await storeChild.reset();
       window.location.reload();
     });
   },
--- a/js/lightbeam.js
+++ b/js/lightbeam.js
@@ -142,11 +142,12 @@ const lightbeam = {
   resetData() {
     const resetData = document.getElementById('reset-data-button');
     resetData.addEventListener('click', async () => {
-      var confirmation = confirm('Pressing OK will delete all'
+      const confirmation = confirm('Pressing OK will delete all'
        + ' Lightbeam information. Are you sure?');
-      if (confirmation == true)
+      if (confirmation) {
         await storeChild.reset();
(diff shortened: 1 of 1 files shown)


js/lightbeam.js
```diff
@@ -142,7 +142,10 @@ const lightbeam = {
   resetData() {
     const resetData = document.getElementById('reset-data-button');
     resetData.addEventListener('click', async () => {
-      await storeChild.reset();
+      var confirmation = confirm('Pressing OK will delete all'
+       + ' Lightbeam information. Are you sure?');
+      if (confirmation == true)
+        await storeChild.reset();
       window.location.reload();
     });
```
princiya: @joanasouza thanks for the PR. please enclose the `if` 
```
if (confirmation === true) {
 -- await
 --reload
}
```

princiya: `confirm` gives true/false values. please make if statement as `if (confirmation)`

princiya:
I shall create a follow-up issue to add CSS to this confirm dialog box.

js/lightbeam.js
princiya: also, please change `confirmation` to `const`. Thanks :)

joanasouza: Thanks, already applied the changes 😄

jonathanKingston:
Sorry to be pedantic but can you make a new variable  for the confirm
string rather than using string concatenation to get under the line limit.
If that is still too long you could use a template string also.

Thanks for the patch! Let me know if you need any help with anything else.

js/lightbeam.js
```diff
@@ -142,8 +142,12 @@ const lightbeam = {
...
     const resetData = document.getElementById('reset-data-button');
     resetData.addEventListener('click', async () => {
-      await storeChild.reset();
-      window.location.reload();
+      const confirmation = confirm('Pressing OK will delete all'
+       + ' Lightbeam information. Are you sure?');
+      if (confirmation) {
+        await storeChild.reset();
+        window.location.reload();
+      }
```
princiya: @joanasouza as per @jonathanKingston 's we could do it like -
`messagePart1` & `messagePart2`?
confirm(`${m1} ${m2}`);

jonathanKingston:
I'm going to merge this however some follow up work could be:
- Changing to meaningful variable names (rather than start and end)
- Do the concat in the template string (rather than having spaces in the strings):
`${msgDeletion} ${msgConfirm}`

Issue #204, opened 2017-09-28
Increase favicons relative to the circle radius


princiya:
![screen shot 2017-09-28 at 15 42 24](https://user-images.githubusercontent.com/8022693/30966107-585443bc-a475-11e7-9b0a-4650f47e3fd6.png)

In this PR #203, the node circle radius is relative to the no. of third party websites corresponding to a first party. We now need to increase the favicon size too, relative to this circle radius.

PR #220, opened 2017-10-04
Increase favicon size relative to the circle radius

Bilie:
Hi

This is aPR for issue [#204](https://github.com/mozilla/lightbeam-we/issues/204)
Looking forward to your feedback!

### DIFF
--- a/src/js/viz.js
+++ b/src/js/viz.js
@@ -23,7 +23,7 @@ const viz = {
     this.collisionRadius = this.collisionRadius * this.scalingFactor;
     this.scale = (window.devicePixelRatio || 1) * this.scalingFactor;
     this.transform = d3.zoomIdentity;
-    this.defaultIcon = this.convertURIToImageData('images/defaultFavicon.svg');
+    this.defaultIcon = this.loadImage('images/defaultFavicon.svg');
 
     this.updateCanvas(width, height);
     this.draw(nodes, links);
@@ -171,12 +171,14 @@ const viz = {
 
       if (node.favicon) {
         this.drawFavicon(node, x, y, radius);
+      } else {
+        this.drawFavicon(node, x, y, this.circleRadius);
       }
     }
   },
 
   getSquare(radius) {
-    const side = Math.sqrt(radius * radius * 1.5);
+    const side = Math.sqrt(radius * radius * 2);
     const offset = side * 0.5;
 
     return {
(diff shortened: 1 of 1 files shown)


princiya:
Thanks for the PR😊. At the moment i am travelling and have poor internet.
Will get back in 2 days.

jonathanKingston:
updates-favicon-lightbeam

One side note is we don't want to request the image again which fixing this might lead to.
Instead pick a max favicon size for `convertURIToImageData` and scale it to the right size when it is being drawn in `drawFavicon`.

src/js/viz.js
```diff
@@ -212,16 +207,28 @@ const viz = {
     });
   },
 
-  async drawFavicon(node, x, y) {
-    const offset = this.getSquare().offset,
+  async drawFavicon(node, x, y, radius) {
```
princiya: @Bilie thanks for this PR. However, I am not convinced about how moving the code to `drawFavicon` function changes anything? Please correct me if I am missing anything obvious.
Moreover, now we have lost the caching functionality and on each call to `drawFavicon`, `imageCanvas` is being set.
Here is the [PR](https://github.com/mozilla/lightbeam-we/pull/201/files) where we implemented the caching functionality. With the moving of the code into the `drawFavicon` function, we now have come back to the non-caching version.

jonathanKingston: From what I can tell this PR caches the network request which is great however In think we can save ourselves from re-rendering the favicon on every load twice.  Currently you are returning the canvas object then scaling it down. I personally think it makes more sense for this to be done inside `convertURIToImageData` as @princiya is mentioning and return the image data of a 64px icon perhaps, then use this to scale down to the right size on the main canvas.

So I think something like the following code can be in  `convertURIToImageData`:
```
    const imageContext = node.imageCanvas.getContext('2d');
    const side = this.getSquare(radius).side * this.scale;

    node.imageCanvas.width = side;
    node.imageCanvas.height = side;
    imageContext.fillStyle = this.canvasColor;
    imageContext.fillRect(0, 0, this.maxSideSize, this.maxSideSize);
    imageContext.drawImage(node.image, 0, 0, this.maxSideSize, this.maxSideSize);

   return imageContext.getImageData(0, 0, this.maxSideSize, this.maxSideSize);
```

Then in `drawIcon` just scale that icon down to the right size.

Bilie:
Hi, thank you very much for all your feedback! I just updated the PR, even though I am not sure this is what was expected. I have tried scaling down the image within `putImageData`, but then the image was just cropped. The only way I found scaling the image is when it is actually drawn with `drawImage`. The image now should be loaded once, but then the image is redrawn on the canvas in the correct size every time nodes are updated.

jonathanKingston:
This appears not to have the perf issues before thanks @Bilie!

Issue #190, opened 2017-08-31
Add button in First Run page to open Lightbeam tab


biancadanforth:
PR #186 is a good basic set up for Lightbeam's first run page (`first-run.html`), but it would be helpful to add a button on this page to make it easier for users to open the Lightbeam page (`index.html`) for the first time.

maqboolkhan:
hi i want to work on it kindly guide me what is the requirement as i could not understand properly
Thanks

jonathanKingston:
There is already a PR outstanding for this work, I think we are good here. I just haven't checked it over yet.

maqboolkhan:
ok thanks

PR #222, opened 2017-10-05
190-added-in-link-to-index

darrylpargeter:
Adds in feature requested in #190 

I have added a button that takes the user from the first-run.html page to index.html
# Starting State:
lightbeam_button_no_hover

# Hovering State:
lightbeam_button_with_hover

### DIFF
--- a/src/css/first-run.css
+++ b/src/css/first-run.css
@@ -15,11 +15,11 @@ h1 {
 }
 
 .center { text-align: center; }
-img.center, button.center { display: block; margin: 0 auto; }
+img.center, button.center, a.center { display: block; margin: 0 auto; }
 
 img#lightbeam-logo { margin-bottom:-18px; }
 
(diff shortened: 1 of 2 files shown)


jonathanKingston:
Hey this looks great!

Can you change the "clicking the link below" to "using the 'Get Started' below".
Then change "Click Here To Get Started" to "Get Started"

Rationale is, we should allow keyboard navigation. I'm less worried about the browser action as the primary action is clicking but if you can think of better wording to that would be great.

darrylpargeter:
Hey @jonathanKingston I have update the PR with though changes.

jonathanKingston:
LGTM thanks @darrylpargeter, currently we don't display this page as we didn't have the code in place to run on install vs update. So I disabled it, will raise a follow up bug.

darrylpargeter:
Hey @jonathanKingston  when you have created that ticket would it be possible to tackle it?

PR #231, opened 2017-10-18
Use consistent braces for arrow functions

### LINKED ISSUE #64: Use consistent braces for arrow functions
From #62

> Please add a follow up bug to make this use braces. Having the implied return may in future cause issues as this method returns a promise...blocking requests for example would delay.

```
(response) => this.sendThirdParty(response),
```

IF you can find an Eslint rule that would be great too.

epq:
Fixes #64. 
* Added ESLint rule arrow-body-style
* Added braces where needed

### DIFF
--- a/.eslintrc.json
+++ b/.eslintrc.json
@@ -120,6 +120,10 @@
             "functions": "always",
             "keywords": "always"
           }
+        ],
+        "arrow-body-style": [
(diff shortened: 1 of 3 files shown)


jonathanKingston:
Thanks! Sorry for the delay.

PR #234, opened 2017-11-19
Remove manual tick

princiya:
In this PR:

- removed manual tick
- improved alpha values

The explanation about the improved alpha is in this PR - #233


jonathanKingston:
Sorry there was a confusion on what this patch was. This looks good to me! Thanks again and as always!

